PR #2, opened 2023-07-27
feat: add prompt model

alangsto:
## [MST-2033](https://2u-internal.atlassian.net/browse/MST-2033)

Add model to store relationship between prompt text and course ID. Added class method to retrieve prompt text given a course ID.


learning_assistant/models.py
```diff
@@ -1,3 +1,39 @@
...
+
+
+class CoursePrompt(TimeStampedModel):
+    """
+    This model represents a mapping between a particular course ID and a text prompt associated with the course ID.
+
+    .. no_pii: This model has no PII.
+    """
+
+    # course ID with which the text prompt is associated with
```
MichaelRoytman, line 15: Nit. You say "with" twice here.

```suggestion
    # course ID with which the text prompt is associated
```

learning_assistant/models.py
```diff
@@ -1,3 +1,39 @@
...
+    course_id = models.CharField(max_length=255, db_index=True, unique=True)
+
+    # text prompt, that may contain course related information
+    prompt = models.TextField(blank=True)
+
+    class Meta:
```
MichaelRoytman, line 21: Is there a reason you included the Meta class? I believe that the values you set for `db_table` and `verbose_name` should be the default values that Django will compute and user under-the-hood.

learning_assistant/models.py
```diff
@@ -1,3 +1,39 @@
...
+
+class CoursePrompt(TimeStampedModel):
+    """
+    This model represents a mapping between a particular course ID and a text prompt associated with the course ID.
+
+    .. no_pii: This model has no PII.
+    """
+
+    # course ID with which the text prompt is associated with
+    course_id = models.CharField(max_length=255, db_index=True, unique=True)
```
MichaelRoytman, line 16: I think we should use a `CourseKeyField` here, if we can. The [Opaque Keys (Locators)](https://github.com/openedx/edx-platform/wiki/Opaque-Keys-(Locators)) wiki has a good summary of the problems this field solves. What do you think?

If we ever need to manipulate or extract information from the course key (e.g. get the associated organization), using a course key field will make that a lot easier, for example.

alangsto: I'll take a look - I know we used a CharField for exams here: https://github.com/edx/edx-exams/blob/c3bdd6c76da2caaa53076798d7e5cfa9cbab9276/edx_exams/apps/core/models.py#L91.

I'm pretty open to either - and I think the option to have more info is better!

learning_assistant/models.py
alangsto: I had seen it included for most other models in repos that we own - I can remove it though!

learning_assistant/models.py
MichaelRoytman: Interesting! I wonder why.

I do see `opaque_keys` used throughout our various repositories (including IDAs and plugins), so I can't think of a good reason not to.

https://github.com/search?q=org%3Aopenedx%20opaque_keys&type=code

learning_assistant/models.py
MichaelRoytman: Hm. You could try removing the Meta class, recreating the migration, running the migration, and see if the effect is the same in the database? I'm usually in favor of not having the code if the framework will do it the same way anyway.

Could it be that those other models have app names that we don't want in the table name? For example, I see `edx_proctoring` models have a Meta class with `db_table`, but I think that's to avoid a table name of `edxproctoring_proctoredexam`, so we set it to `proctoring_proctoredexam`. Maybe we just copied it from `edx-proctoring`?

alangsto: Yeah, it's probably something we've copied over time. It looks fine without the meta class 👍

MichaelRoytman (commented):
This looks good to me! 🚀

MichaelRoytman (approved):

PR #3, opened 2023-08-01
feat: add endpoint to retrieve chat response

alangsto:
## [MST-2035](https://2u-internal.atlassian.net/browse/MST-2035).

This PR contains support for an endpoint that accepts POST requests containing a message list. Access to the endpoint is gated by user enrollment or staff status, a course waffle flag, and an additional Django setting that defines a chat completion API URL. 

If all requirements for access to the endpoint are met, the message list is passed on to the chat completion API. The response from the chat completion API is then passed through as the response from the endpoint defined in the PR.

### DIFF
--- a/learning_assistant/constants.py
+++ b/learning_assistant/constants.py
@@ -1,0 +1,10 @@
+"""
+Constants for the learning_assistant app.
+"""
+# Pulled from edx-platform. Will correctly capture both old- and new-style
+# course ID strings.
+INTERNAL_COURSE_KEY_PATTERN = r'([^/+]+(/|\+)[^/+]+(/|\+)[^/?]+)'
+
+EXTERNAL_COURSE_KEY_PATTERN = r'([A-Za-z0-9-_:]+)'
+
+COURSE_ID_PATTERN = rf'(?P<course_id>({INTERNAL_COURSE_KEY_PATTERN}|{EXTERNAL_COURSE_KEY_PATTERN}))'
--- a/learning_assistant/serializers.py
+++ b/learning_assistant/serializers.py
@@ -1,0 +1,22 @@
+"""
+Serializers for the learning-assistant API.
+"""
+from rest_framework import serializers
+
+
+class MessageSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    """
+    Serializer for a message.
+    """
+
+    role = serializers.CharField(required=True)
+    content = serializers.CharField(required=True)
+
+    def validate_role(self, value):
+        """
+        Validate that role is one of two acceptable values.
+        """
+        valid_roles = ['user', 'assistant']
+        if value not in valid_roles:
+            raise serializers.ValidationError('Must be valid role.')
+        return value
--- a/learning_assistant/utils.py
+++ b/learning_assistant/utils.py
@@ -1,0 +1,46 @@
+"""
+Utils file for learning-assistant.
+"""
+import logging
+
+import requests
+from django.conf import settings
+from requests.exceptions import ConnectTimeout
+from rest_framework import status as http_status
+
+log = logging.getLogger(__name__)
+
+
+def get_chat_response(message_list):
+    """
+    Pass message list to chat endpoint, as defined by the CHAT_COMPLETION_API setting.
+    """
+    completion_endpoint = getattr(settings, 'CHAT_COMPLETION_API', None)
+    if completion_endpoint:
+        headers = {'Content-Type': 'application/json'}
+        connect_timeout = getattr(settings, 'CHAT_COMPLETION_API_CONNECT_TIMEOUT', 1)
+        read_timeout = getattr(settings, 'CHAT_COMPLETION_API_READ_TIMEOUT', 10)
+
+        try:
+            response = requests.post(
+                completion_endpoint,
+                headers=headers,
+                data=message_list,
+                timeout=(connect_timeout, read_timeout)
+            )
+            chat = response.json()
+            response_status = response.status_code
+        except (ConnectTimeout, ConnectionError) as e:
+            error_message = str(e)
+            connection_message = 'Failed to connect to chat completion API.'
+            log.error(
+                '%(connection_message)s %(error)s',
+                {'connection_message': connection_message, 'error': error_message}
+            )
+            chat = connection_message
+            response_status = http_status.HTTP_502_BAD_GATEWAY
+    else:
+        response_status = http_status.HTTP_404_NOT_FOUND
+        chat = 'Completion endpoint is not defined.'
+
+    return response_status, chat
--- a/learning_assistant/views.py
+++ b/learning_assistant/views.py
@@ -1,0 +1,100 @@
+"""
+V1 API Views.
+"""
+import logging
+
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from rest_framework import status as http_status
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+try:
+    from common.djangoapps.course_modes.models import CourseMode
+    from common.djangoapps.student.models import CourseEnrollment
+    from lms.djangoapps.courseware.access import get_user_role
+    from lms.djangoapps.courseware.toggles import learning_assistant_is_active
+except ImportError:
+    # If the waffle flag is false, the endpoint will force an early return.
+    learning_assistant_is_active = False
+
+from learning_assistant.api import get_prompt_by_course_id
+from learning_assistant.serializers import MessageSerializer
+from learning_assistant.utils import get_chat_response
+
+log = logging.getLogger(__name__)
+
+
+class CourseChatView(APIView):
+    """
+    View to retrieve chat response.
+    """
+
+    authentication_classes = (SessionAuthentication, JwtAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request, course_id):
+        """
+        Given a course ID, retrieve a chat response for that course.
+
+        Expected POST data: {
+            [
+                {'role': 'user', 'content': 'What is 2+2?'},
+                {'role': 'assistant', 'content': '4'}
+            ]
+        }
+        """
+        if not learning_assistant_is_active(course_id):
+            return Response(
+                status=http_status.HTTP_403_FORBIDDEN,
+                data={'detail': 'Learning assistant not enabled for course.'}
+            )
+
+        # If user does not have a verified enrollment, or is not staff, they should not have access
+        user_role = get_user_role(request.user, course_id)
+        enrollment_object = CourseEnrollment.get_enrollment(request.user, course_id)
+        enrollment_mode = enrollment_object.mode if enrollment_object else None
+        if (
+            (enrollment_mode not in CourseMode.VERIFIED_MODES)
+            and user_role not in ('staff', 'instructor')
+        ):
+            return Response(
+                status=http_status.HTTP_403_FORBIDDEN,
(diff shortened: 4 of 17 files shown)


learning_assistant/views.py
```diff
@@ -0,0 +1,101 @@
...
+
+        log.info(
+            'Attempting to retrieve chat response for user_id={user_id} in course_id={course_id}'.format(
+                user_id=request.user.id,
+                course_id=course_id
+            )
+        )
+        status_code, message = get_chat_response(message_setup + message_list)
+
+        return Response(status=status_code, data=message)
```
alangsto, line 101: I'm not sure if it's the "correct" thing to pass through the status_code, but I want any HTTP errors to be surfaced

learning_assistant/views.py
```diff
@@ -0,0 +1,101 @@
...
+            )
+
+        prompt_text = CoursePrompt.get_prompt_by_course_id(course_id)
+        if not prompt_text:
+            return Response(
+                status=http_status.HTTP_404_NOT_FOUND,
+                data={'detail': 'Learning assistant not enabled for course.'}
+            )
+
+        message_list = request.data
```
alangsto, line 76: The request structure for messages is different than what is currently being done in https://github.com/kalebjdavenport/edx-chatbot/blob/c07f1f879e92488d606e86dfb40d1f9c2ff10d48/src/widgets/Xpert.jsx#L70-L74, but I think there is unnecessary nesting being done in the implementation.

learning_assistant/utils.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    completion_endpoint = getattr(settings, 'CHAT_COMPLETION_API', None)
+    if completion_endpoint:
+        headers = {'Content-Type': 'application/json'}
+        connect_timeout = getattr(settings, 'CHAT_COMPLETION_API_CONNECT_TIMEOUT', 1)
+        read_timeout = getattr(settings, 'CHAT_COMPLETION_API_READ_TIMEOUT', 10)
+
+        try:
+            response = requests.post(
```
alangsto, line 26: I'm not sure if this connection point should be in its own client class, but put it in a utils file for now.

learning_assistant/views.py
```diff
@@ -0,0 +1,101 @@
...
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+try:
+    from lms.djangoapps.courseware.access import get_user_role
+    from lms.djangoapps.courseware.toggles import learning_assistant_is_active
+    from common.djangoapps.course_modes.models import CourseMode
+    from common.djangoapps.student.models import CourseEnrollment
+except ImportError:
```
alangsto, line 18: I'm having trouble recreating this import error during testing, because of how I'm mocking the imports. I tried using `patch.dict` to mock each edx-platform module, but because of how the test_views.py file works, that patch isn't actually mocking the modules being used in views.py.

tests/test_views.py
```diff
@@ -0,0 +1,98 @@
...
+
+from learning_assistant.models import CoursePrompt
+from tests import LoggedInTestCase
+
+
+class CourseChatViewTests(LoggedInTestCase):
+    """
+    Test for the CourseChatView
+    """
+    sys.modules['lms.djangoapps.courseware.access'] = MagicMock()
```
alangsto, line 18: When I put these mocks in the setUp function, the modules weren't being mocked properly.

learning_assistant/serializers.py
```diff
@@ -0,0 +1,40 @@
...
+    def validate_role(self, value):
+        """
+        Validate that role is one of two acceptable values.
+        """
+        valid_roles = ['user', 'assistant']
+        if value not in valid_roles:
+            raise serializers.ValidationError('Must be valid role.')
+        return value
+
+    def update(self, instance, validated_data):
```
alangsto, line 24: I'm not sure if this is the right approach, but in order to use the BaseSerializer I need to implement both `update` and `create`, even if they're not being used (otherwise pylint will complain)

tests/test_views.py
MichaelRoytman: Why do we need to mock these? Are you able to let the imports fail and then mock out the imported functions? Or does that not work if the import statement fails?

learning_assistant/views.py
MichaelRoytman: Do you think it's worth testing this case? I think having good test coverage is important, but is it important that we test that the Django application functions a certain way outside of the `edx-platform`? It's designed to be installed into the platform. It's not going to be functional regardless, so I would argue this is a case we can leave uncovered, especially if we're getting into writing tests that are more difficult to understand (i.e. mocking `sys.modules`) or for cases that are difficult to test and are not vital to the behavior of the application. What do you think?

learning_assistant/utils.py
```diff
@@ -0,0 +1,42 @@
...
+            )
+            chat = response.json()
+            response_status = response.status_code
+        except (ConnectTimeout, ConnectionError) as e:
+            error_message = str(e)
+            log.error('Failed to connect to chat completion API. %(error)s', {'error': error_message})
+            chat = error_message
+            response_status = http_status.HTTP_502_BAD_GATEWAY
+    else:
+        response_status = http_status.HTTP_404_NOT_FOUND
```
MichaelRoytman, line 39: I'm not sure if `404` is the most appropriate HTTP status here, as it implies a client-side error, and it's an issue of server-side configuration. Maybe something from the `500` class would be more appropriate?

learning_assistant/utils.py
```diff
@@ -0,0 +1,42 @@
...
+                headers=headers,
+                data=message_list,
+                timeout=(connect_timeout, read_timeout)
+            )
+            chat = response.json()
+            response_status = response.status_code
+        except (ConnectTimeout, ConnectionError) as e:
+            error_message = str(e)
+            log.error('Failed to connect to chat completion API. %(error)s', {'error': error_message})
+            chat = error_message
```
MichaelRoytman, line 36: Do we want to send back a Python error as the chat response in this case?

I imagine we do not want that displayed in the UI or visible in the network tab. What's our plan for handling this case in the UI?

learning_assistant/views.py
```diff
@@ -0,0 +1,100 @@
...
+            ]
+        }
+        """
+        if not learning_assistant_is_active(course_id):
+            return Response(
+                status=http_status.HTTP_403_FORBIDDEN,
+                data={'detail': 'Learning assistant not enabled for course.'}
+            )
+
+        # get user does not have a verified enrollment, or is not staff, they should not have access
```
MichaelRoytman, line 54: I think?

```suggestion
        # If user does not have a verified enrollment, or is not staff, they should not have access
```

learning_assistant/views.py
```diff
@@ -0,0 +1,100 @@
...
+        if (
+            (enrollment_mode not in CourseMode.VERIFIED_MODES)
+            and user_role not in ('staff', 'instructor')
+        ):
+            return Response(
+                status=http_status.HTTP_403_FORBIDDEN,
+                data={'detail': 'Must be staff or have valid enrollment.'}
+            )
+
+        prompt_text = CoursePrompt.get_prompt_by_course_id(course_id)
```
MichaelRoytman, line 67: Ideally, I think we would want to have a Python API module to avoid directly interfacing with the model class. But we can add that later.

learning_assistant/serializers.py
MichaelRoytman: That's interesting. The DRF documentation has a serializer with just the serializer fields declared. What's the Pylint rule that's failing?

MichaelRoytman (commented):
I got swamped today and didn't have time to finis this review, but I wanted to post the comments I do have. I'll return to this tomorrow.

learning_assistant/views.py
alangsto: I agree! This is intended to be an edx-platform plugin, so I think it makes sense to assume that the edx-platform modules would be available.

tests/test_views.py
alangsto: I need to mock them so that an `ImportError` isn't thrown when running the tests. If I try to mock the imported functions, that also fails because the module isn't available. 

I tried to use `patch.dict` to update `sys.modules`, but because the tests are hitting an endpoint and not testing a specific function that can be imported, that method of patching doesn't work.

learning_assistant/constants.py
```diff
@@ -0,0 +1,8 @@
+"""
+Constants for the learning_assistant app.
+"""
+INTERNAL_COURSE_KEY_PATTERN = r'([^/+]+(/|\+)[^/+]+(/|\+)[^/?]+)'
+
+EXTERNAL_COURSE_KEY_PATTERN = r'([A-Za-z0-9-_:]+)'
+
+COURSE_ID_PATTERN = rf'(?P<course_id>({INTERNAL_COURSE_KEY_PATTERN}|{EXTERNAL_COURSE_KEY_PATTERN}))'
```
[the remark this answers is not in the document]
alangsto: These are copied from https://github.com/edx/edx-exams/blob/main/edx_exams/apps/core/constants.py. I added the same comment from the edx-exams file to make this more clear!

learning_assistant/views.py
alangsto: I believe that's true in the case of actually modifying the object. I feel comfortable importing the model in the case that it's only used to retrieve data

learning_assistant/serializers.py
alangsto: Maybe it would make more sense to ignore the pylint errors for now, as none of the abstract classes are being used for the serializer right now.

learning_assistant/utils.py
alangsto: I'm not seeing anything in the 500s that would really describe this situation. The reason I chose 404 is because there's no resource associated with the endpoint even though the endpoint itself is valid. This seemed like the most appropriate status code for this case.

learning_assistant/serializers.py
MichaelRoytman: I think that's valid. I see that pattern in [registrar](https://github.com/openedx/registrar/blob/master/registrar/apps/grades/serializers.py#L7) as well.

learning_assistant/utils.py
MichaelRoytman: I thought 404 is referring to the requested resource (i.e. the chatbot feature for this course, in this case) not existing. Like, if the waffle flag for the course was disabled, I'd expect a 400 class error. The 400 class of errors is telling the client it did something wrong (e.g. requested a page that doesn't exist or an object or resource that doesn't exist), whereas the reality is that the server is misconfigured in this case. It doesn't seem like a client issue to me. A request to this endpoint is not going to work for any request to this endpoint if that setting is missing. Maybe a `501 Not Implemented` would make sense, but I can't really tell if that's referring purely to the request method not being supported or _something_ in the endpoint not being supported, which would be closer to this case.

A 404 is fine with me. I'm not an expert on status codes, so I don't want to get hung up on it!

learning_assistant/views.py
MichaelRoytman: I don't think that's true, based on our [OEP](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0049-django-app-patterns.html#api-py).

> An app’s Django models and other internal data structures should not be exposed via its Python APIs (unless performance requires it).

If we shouldn't expose or export a model from a Python API, I don't think we should import a model directly either. I don't think we want to expose the model structure or the functions of the model because it makes it harder to maintain a consistent, well-supported interface for reading from and writing to the model, particularly if we want to make changes to the model or its functions.

Let me know if you read that differently, but it's definitely up to you!

alangsto: I see what you're saying, I can update to move this to an API

MichaelRoytman (approved):
Thanks for addressing my feedback!

PR #7, opened 2023-08-09
fix: course id should be converted to course key for edx-platform functions

alangsto:
## [MST-2045](https://2u-internal.atlassian.net/browse/MST-2045)

The edx-platform functions imported require a course key to be passed in, not a course id. Passing in the course ID was causing a 500 error when making calls to the endpoint. I tested locally and the endpoint is able to make all the required checks prior to forwarding the request to the chat completion endpoint. 

As part of testing, I also needed to set up a course prompt, so I included the ability to add a course prompt via django admin to this PR.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,10 @@ Change Log
 
 .. There should always be an "Unreleased" section for changes pending release.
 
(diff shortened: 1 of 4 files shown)


schenedx (approved):
👍

PR #15, opened 2023-08-25
fix: remove deserialization of jsonfield

alangsto:
The JSONField does not require deserialization, as the data is represented in its Python native format.

Tested locally and confirmed that the new json prompt field is being handled correctly and appended to the list of messages as expected.


spencertiberi (approved):
So the change is to not even think of it as JSON and just use dictionaries, it seems? Cool

alangsto:
@spencertiberi, yes, I believe the JSONField stores it as JSON in the DB, but when it's accessed in python you can treat that field as a native python object. In this case we can just treat it as a list.

PR #21, opened 2023-09-11
feat: use reduced message to avoid maxing out tokens

alangsto:
## [MST-2134](https://2u-internal.atlassian.net/browse/MST-2134)

To avoid going over the maximum amount of tokens allowed, we should reduce the user message list and send only the most recent messages that do not go over the number of allotted tokens.

### DIFF
--- a/learning_assistant/utils.py
+++ b/learning_assistant/utils.py
@@ -12,7 +12,42 @@ from rest_framework import status as http_status
 log = logging.getLogger(__name__)
 
 
-def get_chat_response(message_list):
+def _estimated_message_tokens(message):
+    """
+    Estimates how many tokens are in a given message.
+    """
+    chars_per_token = 3.5
+    json_padding = 8
+
+    return int((len(message) - message.count(' ')) / chars_per_token) + json_padding
+
+
+def get_reduced_message_list(system_list, message_list):
+    """
+    If messages are larger than allotted token amount, return a smaller list of messages.
+    """
+    total_system_tokens = sum(_estimated_message_tokens(system_message['content']) for system_message in system_list)
+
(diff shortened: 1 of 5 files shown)


learning_assistant/utils.py
```diff
@@ -12,7 +12,42 @@
...
+    """
+    Estimates how many tokens are in a given message.
+    """
+    chars_per_token = 3.5
+    json_padding = 8
+
+    return int((len(message) - message.count(' ')) / chars_per_token) + json_padding
+
+
+def get_reduced_message_list(system_list, message_list):
```
alangsto, line 25: This was largely taken from https://github.com/2uinc/ai-chatbot-backend/blob/74faef96cc3732e1369d3ad5adeae9cdd79d0bc8/xpert/pipelines/utils.py#L173. I modified a few things within the while loop (like moving a break and adding it to the outer conditional) and changing the append + reverse to an insert for the `new_message_list`

learning_assistant/utils.py
```diff
@@ -12,7 +12,42 @@
...
+    """
+    If messages are larger than allotted token amount, return a smaller list of messages.
+    """
+    total_system_tokens = sum(_estimated_message_tokens(system_message['content']) for system_message in system_list)
```
alangsto, line 29: This doesn't currently account for if the total_system_tokens is greater than the number of allowed tokens, but I couldn't find anything in https://github.com/2uinc/ai-chatbot-backend on the same codepath as this function that checks for that. I think it's fine as is but probably worth noting in the discovery for https://2u-internal.atlassian.net/browse/MST-2122.

Zacharis278 (approved):

PR #37, opened 2023-11-28
feat: expose functionality to extract unit content

alangsto:
## [COSMO-70](https://2u-internal.atlassian.net/browse/COSMO-70)

In order to provide unit specific content to a prompt template, the content needs to be extracted from a given block. Pieces of this code have been taken from the ai_aside repository, specifically the `text_utils.py` file. 

Content from a block is retrieved using a pre-order traversal, meaning that all content will be returned in the order a learner would be viewing the content.

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -1,7 +1,16 @@
 """
 Library for the learning_assistant app.
 """
+from learning_assistant.constants import ACCEPTED_CATEGORY_TYPES, CATEGORY_TYPE_MAP
 from learning_assistant.models import CoursePrompt
+from learning_assistant.platform_imports import (
+    block_get_children,
+    block_leaf_filter,
+    get_single_block,
+    get_text_transcript,
+    traverse_block_pre_order,
+)
+from learning_assistant.text_utils import html_to_text
 
 
 def get_deserialized_prompt_content_by_course_id(course_id):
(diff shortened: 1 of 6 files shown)


learning_assistant/api.py
```diff
@@ -23,3 +32,69 @@ def get_setup_messages(course_id):
...
+        return text
+
+    if category == 'video':
+        transcript = get_text_transcript(child)  # may be None
+        return transcript
+
+    return None
+
+
+def _leaf_filter(block):
```
alangsto, line 55: This is a custom leaf filter to exclude blocks that are not video or html, that way we limit any calculations/processing on nodes that aren't relevant to what we need.

learning_assistant/api.py
```diff
@@ -23,3 +32,69 @@ def get_setup_messages(course_id):
...
+
+def get_block_content(request, user_id, course_id, unit_usage_key):
+    """
+    Public wrapper for retrieving the content of a given block's children.
+
+    Returns
+        length - the cummulative length of a block's children's content
+        items - a list of dictionaries containing the content type and text for each child
+    """
+    block = get_single_block(request, user_id, course_id, unit_usage_key)
```
alangsto, line 96: I'm only using `load_single_xblock` once here on the parent block. I wasn't sure if I need to do that on each child block to check visibility.

learning_assistant/platform_imports.py
```diff
@@ -0,0 +1,47 @@
...
+    from xmodule.video_block.transcripts_utils import get_transcript
+    try:
+        transcript, _, _ = get_transcript(video_block, output_format='txt')
+    except NotFoundError:
+        # some old videos have no transcripts, just accept that reality
+        return None
+    return transcript
+
+
+def get_single_block(request, user_id, course_id, usage_key_string, course=None, will_recheck_access=False):
```
MichaelRoytman, line 22: I don't think we'll ever want this to be `True` because then we won't be checking learner access against to the block. I think it would be better not to expose this function parameter in that case, since we don't want someone to accidentally pass `True`. What do you think?

learning_assistant/api.py
MichaelRoytman: You shouldn't need to, no. `load_single_xblock` should only return the children that the learner has access to, so long as `will_recheck_access=False`.

learning_assistant/api.py
```diff
@@ -23,3 +32,69 @@ def get_setup_messages(course_id):
...
+    category = block.category
+
+    return is_leaf and category in ACCEPTED_CATEGORY_TYPES
+
+
+def _get_children_contents(block):
+    """
+    Given a specific block, return the content type and text of a pre-order traversal of the blocks children.
+    """
+    leaf_nodes = traverse_block_pre_order(block, block_get_children, _leaf_filter)
```
MichaelRoytman, line 69: This is slick!

learning_assistant/api.py
```diff
@@ -23,3 +32,69 @@ def get_setup_messages(course_id):
...
+            length += len(content)
+            items.append({
+                'content_type': CATEGORY_TYPE_MAP.get(category),
+                'content_text': content,
+            })
+
+    return length, items
+
+
+def get_block_content(request, user_id, course_id, unit_usage_key):
```
Zacharis278, line 88: for my own understanding, what calls this?

Zacharis278 (approved):

tests/test_api.py
```diff
@@ -38,3 +82,111 @@ def test_get_setup_messages(self):
...
+                        </The quick brown fox jumps over the lazy dog>'''),
+            FakeChild('video', '03'),
+            FakeChild('unknown', '04')
+        ]
+        self.block = FakeBlock(self.children)
+
+        self.course_id = 'course-v1:edx+test+23'
+
+    @ddt.data(
+        ('video', True),
```
MichaelRoytman, line 114: Should we include `html` here for completeness?

MichaelRoytman (approved):
This is very easy to understand, so I appreciate the thought and care you took in putting this together.

tests/test_api.py
alangsto: Yes, thanks for catching that!

learning_assistant/platform_imports.py
alangsto: Oh, that's a great point! I'll update this function.

learning_assistant/api.py
alangsto: Got it, thank you!

learning_assistant/api.py
alangsto: This function will be used as part of https://2u-internal.atlassian.net/browse/COSMO-73 - essentially whatever code is putting together the prompt will call it.

PR #41, opened 2023-12-13
docs: document system prompt design changes

MichaelRoytman:
### Description

This commit adds documentation for changes to the system prompt design. It includes an [Architecture Decision Record (ADR)](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0019-bp-developer-documentation.html#adrs) detailing our proposed changes to the way we store the system prompt as well as instructions for how to properly modify the system prompt with this new design.

### DIFF
--- a/docs/decisions/0002-system-prompt-design-changes.rst
+++ b/docs/decisions/0002-system-prompt-design-changes.rst
@@ -1,0 +1,96 @@
+2. System Prompt Design Changes
+###############################
+
+Status
+******
+
+**Accepted** *2023-12-13*
+
+Context
+*******
+
+A system prompt in the Learning Assistant context refers to the text-based instructions provided to the large language 
+model (LLM) that describe the objective and proper behavior of the Learning Assistant. This prompt is provided to the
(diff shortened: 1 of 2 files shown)


docs/decisions/0002-system-prompt-design-changes.rst
```diff
@@ -0,0 +1,96 @@
...
+  template will be sent to 2U's Xpert Platform generic chat completion endpoint to be completely rendered.
+* We will remove the `CoursePrompt model`_ following the instructions documented in
+  `Everything About Database Migrations`_.
+
+Consequences
+************
+
+* Changes to the system prompt will require pull requests to the appropriate repository in which the prompt is stored.
+* Anyone with access to the appropriate repository in which the prompt is stored can change the system prompt. Members
+* of the team will no longer require assistance from the SRE team.
```
alangsto, line 56: nit: this should be indented instead of bulleted.

alangsto (approved):
One nit, but otherwise this looks great! Thanks for putting so much detail into both documents.

PR #48, opened 2024-01-04
feat: add LearningAssistantCourseEnabled model to store course overrides

MichaelRoytman:
### Description

**Jira**: [COSMO-134](https://2u-internal.atlassian.net/browse/COSMO-134)

This commit adds a `LearningAssistantCourseEnabled` model to store overrides for whether the Learning Assistant feature is enabled. Although whether the feature is enabled will first be controlled by a `CourseWaffleFlag`, eventually, this model will allow course team members to selectively disable or re-enable the feature themselves.


learning_assistant/models.py
```diff
@@ -30,3 +30,20 @@ def get_json_prompt_content_by_course_id(cls, course_id):
         except cls.DoesNotExist:
             json_prompt_content = None
         return json_prompt_content
+
+
+class LearningAssistantCourseEnabled(TimeStampedModel):
+    """
+    This model stores whether the Learning Assistant is enabled for a particular course ID.
+
+    For now, the purpose of this model is to store overrides added by course team members. By default, the Learning
```
MichaelRoytman, line 39: Let me know if you think I should remove this. It does risk falling out of date when we remove the `CourseWaffleFlag` 🤔.

alangsto: I think it's okay to leave this in! We may want to update the ticket for removing the course waffle flag to include cleaning up this comment, just so we don't forget.

alangsto (approved):

PR #49, opened 2024-01-08
Add ADR for using the edx-platform CourseApp Django app.

MichaelRoytman:
### Description

**Jira**: [COSMO-128](https://2u-internal.atlassian.net/browse/COSMO-128)

This commit adds an ADR to document our decision to use the edx-platform [CourseApp Django app](https://github.com/openedx/edx-platform/tree/master/openedx/core/djangoapps/course_apps) to register a Learning Assistant CourseApp plugin with the platform. The impact of this is that this feature will be exposed via the [CourseApp REST API](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/course_apps/rest_api/v1/views.py#L80), and a card for this feature will appear on the [Pages & Resources page](https://github.com/openedx/frontend-app-course-authoring/tree/master/src/pages-and-resources).

### DIFF
--- a/docs/decisions/0003-courseapp-use.rst
+++ b/docs/decisions/0003-courseapp-use.rst
@@ -1,0 +1,58 @@
+3. CourseApp Use
+################
+
+Status
+******
+
+**Accepted** *2024-01-08*
+
+Context
+*******
+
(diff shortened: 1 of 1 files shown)


alangsto (approved):
Looks great!

PR #50, opened 2024-01-08
feat: add course waffle flag for content integration

alangsto:
## [COSMO-80](https://2u-internal.atlassian.net/browse/COSMO-80)

This pull request adds a course waffle flag that will control whether or not course content is added to the prompt template. 

Because the toggle is defined outside of a function, an import error occurs at runtime if the CourseWaffleFlag class is not mocked out properly.

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -7,6 +7,7 @@ from django.conf import settings
 from django.core.cache import cache
 from edx_django_utils.cache import get_cache_key
 from jinja2 import BaseLoader, Environment
+from opaque_keys.edx.keys import CourseKey
 
 from learning_assistant.constants import ACCEPTED_CATEGORY_TYPES, CATEGORY_TYPE_MAP
 from learning_assistant.models import CoursePrompt
@@ -18,6 +19,7 @@ from learning_assistant.platform_imports import (
     traverse_block_pre_order,
 )
 from learning_assistant.text_utils import html_to_text
+from learning_assistant.toggles import course_content_enabled
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


learning_assistant/toggles.py
```diff
@@ -0,0 +1,28 @@
+"""
+Toggles for learning-assistant app.
+"""
+try:
+    from openedx.core.djangoapps.waffle_utils import CourseWaffleFlag
+except ImportError:
+    # mock out flag to avoid import errors during testing
+    class Flag:
```
alangsto, line 8: Pytest complains if this is not mocked out, as it expects a callable object for CourseWaffleFlag. Because the waffle flag is evaluated at run time, this needs to be mocked in some way.

MichaelRoytman: Does [how ai-aside does it](https://github.com/edx/ai-aside/blob/b2d06406041c09d27ada9dbbd210162fd78d5562/ai_aside/waffle.py) work for you?

alangsto: Yes, that's exactly what I was looking for. Thank you!

tests/test_api.py
```diff
@@ -207,12 +207,17 @@ def test_get_block_content(self, mock_get_children_contents, mock_get_single_blo
...
+        ('This is content.', True),
+        ('', True),
+        ('This is content.', False),
+        ('', False),
     )
+    @ddt.unpack
+    @patch('learning_assistant.toggles._is_learning_assistant_waffle_flag_enabled')
     @patch('learning_assistant.api.get_block_content')
-    def test_render_prompt_template(self, unit_content, mock_get_content):
+    def test_render_prompt_template(self, unit_content, enable_content, mock_get_content, mock_flag_enabled):
```
varshamenon4, line 218: Clarification: could you clarify why there is both enable_content and mock_flag_enabled? Thanks!

learning_assistant/api.py
```diff
@@ -124,7 +126,9 @@ def render_prompt_template(request, user_id, course_id, unit_usage_key):
     Return a rendered prompt template, specified by the LEARNING_ASSISTANT_PROMPT_TEMPLATE setting.
     """
     unit_content = ''
-    if unit_usage_key:
+
+    course_run_key = CourseKey.from_string(course_id)
+    if unit_usage_key and course_content_enabled(course_run_key):
```
varshamenon4, line 131: Nice! Very readable and good variable names. :)

varshamenon4 (approved):
lgtm! Just had a clarifying question.

tests/test_api.py
alangsto: `enable_content` is the value being set for the waffle flag, whereas `mock_flag_enabled` is a mock for the function `_is_learning_assistant_waffle_flag_enabled`. Those names are definitely a bit confusing so I will rename them!

PR #52, opened 2024-01-10
feat: add CourseApp for Learning Assistant feature and register as entrypoint

MichaelRoytman:
### Description

**Jira**: [COSMO-135](https://2u-internal.atlassian.net/browse/COSMO-135)

**NOTE**: This pull request should not be merged until the configuration modal is added to the [Course Authoring MFE](https://github.com/openedx/frontend-app-course-authoring/tree/master/src/pages-and-resources) for this feature. Please see [COSMO-136](https://2u-internal.atlassian.net/browse/COSMO-136) and [COSMO-137](https://2u-internal.atlassian.net/browse/COSMO-137).

This commit adds a concrete implementation of the `CourseApp` ABC from the `CourseApps` [edx-platform Django app](https://github.com/openedx/edx-platform/tree/d2fd7b2e1fd95baef5fe105d79ead517fe16f20d/openedx/core/djangoapps/course_apps). This enables the Learning Assistant to be represented by a card on the Studio Pages & Resources pages. Please see the associated ADR for more details.

This commit also registers this `CourseApp` class as an entrypoint to the `CourseApps` [edx-platfrom Django app](https://github.com/openedx/edx-platform/tree/d2fd7b2e1fd95baef5fe105d79ead517fe16f20d/openedx/core/djangoapps/course_apps). Because the `CourseApps` REST API is a part of the CMS, the Learning Assistant plugin was also added as entrypoint to the CMS.

image
image

### DIFF
--- a/learning_assistant/data.py
+++ b/learning_assistant/data.py
@@ -1,0 +1,15 @@
+"""
+Data classes for the Learning Assistant application.
+"""
+from attrs import field, frozen, validators
+from opaque_keys.edx.keys import CourseKey
+
+
+@frozen
+class LearningAssistantCourseEnabledData:
+    """
+    Data class representing whether Learning Assistant is enabled in a course.
+    """
+
+    course_key: CourseKey = field(validator=validators.instance_of(CourseKey))
+    enabled: bool = field(validator=validators.instance_of(bool))
--- a/learning_assistant/platform_imports.py
+++ b/learning_assistant/platform_imports.py
@@ -77,3 +77,19 @@ def learning_assistant_available(course_key):
     # pylint: disable=import-error, import-outside-toplevel
     from lms.djangoapps.courseware.toggles import learning_assistant_is_active
     return learning_assistant_is_active(course_key)
+
+
+def get_user_role(user, course_key):
+    """
+    Return the role of the user on the edX platform.
+
+    Arguments:
+        * user (User): the user who's role to get
+        * course_key (CourseKey): the key of the course in which to get the user's role
+
+    Returns:
+        * str: the user's role
+    """
+    # pylint: disable=import-error, import-outside-toplevel
+    from lms.djangoapps.courseware.access import get_user_role as platform_get_user_role
+    return platform_get_user_role(user, course_key)
--- a/learning_assistant/plugins.py
+++ b/learning_assistant/plugins.py
@@ -1,0 +1,90 @@
+"""
+Plugins for the Learning Assistant application.
+"""
+# pylint: disable=import-error
+from openedx.core.djangoapps.course_apps.plugins import CourseApp
+
+from learning_assistant import plugins_api
+
+
+class LearningAssistantCourseApp(CourseApp):
+    """
+    A CourseApp plugin representing the Learning Assistant feature.
+
+    Please see the associated ADR for more details.
+    """
+
+    app_id = 'learning_assistant'
+    name = 'Learning Assistant'
+    description = 'Use generative AI to power a Learning Assistant using course content.'
+    documentation_links = {
+        'learn_more_openai': 'https://openai.com/',
+        'learn_more_openai_data_privacy': 'https://openai.com/api-data-privacy',
+    }
+
+    @classmethod
+    def is_available(cls, course_key):
+        """
+        Return a boolean indicating this course app's availability for a given course.
+
+        If an app is not available, it will not show up in the UI at all for that course,
+        and it will not be possible to enable/disable/configure it.
+
+        Args:
+            course_key (CourseKey): Course key for course whose availability is being checked.
+
+        Returns:
+            bool: Availability status of app.
+        """
+        return plugins_api.is_available(course_key)
+
+    @classmethod
+    def is_enabled(cls, course_key):
+        """
+        Return if this course app is enabled for the provided course.
+
+        Args:
+            course_key (CourseKey): The course key for the course you
+                want to check the status of.
+
+        Returns:
+            bool: The status of the course app for the specified course.
+        """
+        return plugins_api.is_enabled(course_key)
+
+    @classmethod
+    def set_enabled(cls, course_key, enabled, user):
+        """
+        Update the status of this app for the provided course and return the new status.
+
+        Args:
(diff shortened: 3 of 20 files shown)


learning_assistant/plugins.py
```diff
@@ -0,0 +1,87 @@
...
+class LearningAssistantCourseApp(CourseApp):
+    """
+    A CourseApp plugin representing the Learning Assistant feature.
+
+    Please see the associated ADR for more details.
+    """
+
+    app_id = 'learning_assistant'
+    name = 'Learning Assistant'
+    description = 'TBD'
```
MichaelRoytman, line 19: TBD: We need to provide a privacy-approved description. DO NOT merge without filling this in.

setup.py
```diff
@@ -131,6 +131,12 @@ def is_requirement(line):
     entry_points={
         'lms.djangoapp': [
             "learning_assistant = learning_assistant.apps:LearningAssistantConfig"
+        ],
+        'cms.djangoapp': [
```
MichaelRoytman, line 135: Because the `CourseApps` API is registered under the CMS, the `Learning Assistant` needs to be registered as a plugin there as well. Otherwise, the plugin is not included in the CMS's `INSTALLED_APPS` list, which causes a runtime error, because the `LearningAssistantCourseApp` plugin refers to the `Learning Assistant` models, which are not available in the CMS.

learning_assistant/plugins_api.py
```diff
@@ -0,0 +1,88 @@
+"""
+Concrete implementations of abstract methods of the CourseApp plugin ABC, for use by the LearningAssistantCourseApp.
+
+Because the LearningAssistantCourseApp plugin inherits from the CourseApp class, which is imported from the
+edx-platform, we cannot test that plugin directly, because pytest will run outside the platform context.
+Instead, the CourseApp abstract methods are implemented here and
+imported into and used by the LearningAssistantCourseApp. This way, these implementations can be tested.
```
Zacharis278, line 7: ahh good idea

learning_assistant/utils.py
```diff
@@ -123,3 +123,16 @@ def get_chat_response(prompt_template, message_list, courserun_id):
         chat = 'Completion endpoint is not defined.'
 
     return response_status, chat
+
+
+def user_role_is_staff(role):
```
Zacharis278, line 128: A bit of forward thinking here. I'm wondering is changing the name to has_permission or something might make us more flexible and account for the future roles and permissions changes coming to the platform. tldr; version is roles would grant permissions and functionality is gated by those permissions, not the roles.

learning_assistant/data.py
```diff
@@ -0,0 +1,15 @@
+"""
+Data classes for the Learning Assistant application.
+"""
+from attrs import field, frozen, validators
+from opaque_keys.edx.keys import CourseKey
+
+
+@frozen
+class LearningAssistantCourseEnabledData:
```
Zacharis278, line 9: what's the data class used for?

learning_assistant/platform_imports.py
```diff
@@ -57,3 +57,39 @@ def get_cache_course_run_data(course_run_id, fields):
...
+    cases, the Learning Assistant may be available and disabled.
+
+    Arguments:
+        * course_key (CourseKey): the course's key
+
+    Returns:
+        * bool: whether the Learning Assistant feature is available
+    """
+    # pylint: disable=import-error, import-outside-toplevel
+    from lms.djangoapps.courseware.toggles import learning_assistant_is_active
```
Zacharis278, line 78: Thinking out loud. Does this create a cms -> learning -> lms import loop and is that a problem in any way?

MichaelRoytman: That's a good question. I don't think so. Where do you think the loop would be? I've tried tracing it out, but I can't think of where there might be a loop currently.

I haven't seen any issues when testing the `CourseApps` API in the CMS or when testing the Pages & Resources page.

Some other plugins are defined in the `lms` directory (e.g. [ProctoringCourseApp](https://github.com/openedx/edx-platform/blob/1f3d23250926751acb80aaba6a6bc3b4d7ac3f42/lms/djangoapps/courseware/plugins.py#L167)), so I'd imagine they would have a similar problem, if I'm understanding you correctly.

learning_assistant/data.py
MichaelRoytman: I was following our [Django App Patterns OEP](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0049-django-app-patterns.html#data-py), so I don't expose a modal instance via the API. I need to return something from `set_learning_assistant_enabled`, because the `CourseApp` ABC requires that the `set_enabled` function implementation returns the value of the toggle.

On second look, I could return a boolean from `set_learning_assistant_enabled` and avoid needing the data class entirely. I think I was treating `set_learning_assistant_enabled` as a create/update operation and was thinking about returning the created/update object, which necessitated the data class. For now, this is a simple model/object, so that could be overkill.

If it feels heavy handed/confusing, let me know. I can return a boolean.

learning_assistant/platform_imports.py
```diff
@@ -57,3 +57,39 @@ def get_cache_course_run_data(course_run_id, fields):
...
+def get_user_role(user, course_key):
+    """
+    Return the role of the user on the edX platform.
+
+    Arguments:
+        * user (User): the user who's role to get
+        * course_key (CourseKey): the key of the course in which to get the user's role
+
+    Returns:
+        * bool: whether the Learning Assistant feature is available
```
alangsto, line 91: NIT: return type is the same as above

alangsto (approved):
One nit, but otherwise this looks good!

learning_assistant/platform_imports.py
Zacharis278: ah I think a 'loop' might be the wrong term here. It's definitely weird to me we're importing from the LMS into the CMS tho. I'd expect shared code to exist in /openedx or /common. I guess if we're already following an established pattern for CourseApps best not to break it however.

learning_assistant/data.py
Zacharis278: makes sense to me 👍

Zacharis278 (approved):

PR #53, opened 2024-01-12
Remove references to the CoursePrompt model in preparation for its removal.

MichaelRoytman:
### Description

**Jira**: [COSMO-120](https://2u-internal.atlassian.net/browse/COSMO-120)

This commit removes references to the `CoursePrompt` model. Because we no longer have plans to store system prompts in a model, we plan to remove this model and drop the associated table. In preparation for this step, we need to remove references to the model.


alangsto (approved):
🚢

PR #60, opened 2024-01-25
feat: return 403 if Learning Assistant feature toggle or instructor toggle disabled

MichaelRoytman:
### Description

**Jira**: [COSMO-161](https://2u-internal.atlassian.net/browse/COSMO-161)

This commit modifies the Learning Assistant access logic.

Previously, the chat view would return a 403 response code if the `learning_assistant_available` function returned False (i.e. the `COURSEWARE_LEARNING_ASSISTANT` `CourseWaffleFlag` was disabled).

Now, the chat view will return a 403 response code if either the `learning_assistant_available` function returns False or if there is a False override in the `LearningAssistantCourseEnabled` model corresponding to the course in which the chat is being requested.

### DIFF
--- a/learning_assistant/platform_imports.py
+++ b/learning_assistant/platform_imports.py
@@ -57,3 +57,23 @@ def get_cache_course_run_data(course_run_id, fields):
     # pylint: disable=import-error, import-outside-toplevel
     from openedx.core.djangoapps.catalog.utils import get_course_run_data
     return get_course_run_data(course_run_id, fields)
+
+
+def learning_assistant_available(course_key):
+    """
+    Return whether the Learning Assistant is available in the course represented by the course_key.
+
+    Note that this may be different than whether the Learning Assistant is enabled in the course. The value returned by
+    this fuction represents whether the Learning Assistant is available in the course and, perhaps, whether it is
+    enabled. Course teams can disable the Learning Assistant via the LearningAssistantCourseEnabled model, so, in those
+    cases, the Learning Assistant may be available and disabled.
+
+    Arguments:
+        * course_key (CourseKey): the course's key
+
+    Returns:
+        * bool: whether the Learning Assistant feature is available
+    """
+    # pylint: disable=import-error, import-outside-toplevel
+    from lms.djangoapps.courseware.toggles import learning_assistant_is_active
+    return learning_assistant_is_active(course_key)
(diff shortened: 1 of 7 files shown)


learning_assistant/platform_imports.py
```diff
@@ -57,3 +57,23 @@ def get_cache_course_run_data(course_run_id, fields):
...
+
+    Arguments:
+        * course_key (CourseKey): the course's key
+
+    Returns:
+        * bool: whether the Learning Assistant feature is available
+    """
+    # pylint: disable=import-error, import-outside-toplevel
+    from lms.djangoapps.courseware.toggles import learning_assistant_is_active
+    return learning_assistant_is_active(course_key)
```
alangsto, line 79: To double check - is the idea that we will eventually get rid of this waffle flag, so that only the toggle controls the visibility of the learning assistant?

learning_assistant/views.py
```diff
@@ -15,12 +15,10 @@
     from common.djangoapps.course_modes.models import CourseMode
     from common.djangoapps.student.models import CourseEnrollment
     from lms.djangoapps.courseware.access import get_user_role
-    from lms.djangoapps.courseware.toggles import learning_assistant_is_active
 except ImportError:
-    # If the waffle flag is false, the endpoint will force an early return.
-    learning_assistant_is_active = False
```
alangsto, line 21: Thanks for cleaning this up!

learning_assistant/platform_imports.py
MichaelRoytman: In an ideal world, I think we would convert the waffle flag to a Django setting, since the flags are meant to be temporary. It's not that the role of the flag goes away, it just becomes a setting; we would still have to check the value of the setting.

My assumption when I wrote this was that we'd have a Django setting controlling whether Xpert is on or off platform-wide. Then, the toggle model would be a way for select courses to turn the Xpert feature off (or back on). It's as if we split the current `CourseWaffleFlag` into two pieces - the platform "flag" (i.e. the setting) and the course "flag" override (i.e. the model). But I wasn't intending to replace the role of the `CourseWaffleFlag` with the toggle model, because the toggle model doesn't support the platform-wide toggle behavior.

Although, maybe a setting counts as "Xpert-specific code" in the platform, and we don't want to do that 🤔. I think we'd read the setting only from the plugin, and the setting wouldn't be read in the actual platform, so I think it should be fine. Are you thinking the latter? I could see how that would throw a wrench in this.

alangsto: Ah yes, I just needed the reminder that this is being converted to a django setting. The setting could be in edx-internal, and that way we would avoid Xpert-specific code in edx-platform. 

Thank you for the explanation!

alangsto (approved):

PR #63, opened 2024-01-30
feat: add GET endpoint to retrieve whether Learning Assistant is enabled

MichaelRoytman:
### Description

Jira: [COSMO-161](https://2u-internal.atlassian.net/browse/COSMO-161)

This commit adds a new `GET` endpoint to retrieve whether the Learning Assistant feature is enabled in a course, given a course ID. This endpoint will introspect all data that is relevant to determining whether the feature is enabled (i.e. the `courseware.learning_assistant` `CourseWaffleFlag` and the `LearningAssistantCourseEnabled` model).

This endpoint was added so that the frontend can determine whether or not to show the Learning Assistant feature without requiring that the [edx-platform](https://github.com/openedx/edx-platform) read from the `LearningAssistantCourseEnabled` model in this plugin.


learning_assistant/urls.py
```diff
@@ -4,14 +4,19 @@
...
 from learning_assistant.constants import COURSE_ID_PATTERN
-from learning_assistant.views import CourseChatView
+from learning_assistant.views import CourseChatView, LearningAssistantEnabledView
 
 app_name = 'learning_assistant'
 
 urlpatterns = [
     re_path(
-        fr'learning_assistant/v1/course_id/{COURSE_ID_PATTERN}',
+        fr'learning_assistant/v1/course_id/{COURSE_ID_PATTERN}$',
```
MichaelRoytman, line 13: This `$` tells Django to match the regular expression fully; see [documentation](https://docs.djangoproject.com/en/5.0/ref/urls/#re-path). Without this symbol, a call to `learning_assistant/v1/course_enabled/<course_id>/enabled` would be matched by this regular expression and map to the wrong view.

learning_assistant/views.py
```diff
@@ -91,3 +91,41 @@ def post(self, request, course_id):
...
+        Given a course ID, retrieve whether the Learning Assistant is enabled for the corresponding course.
+
+        The response will be in the following format.
+
+            {'enabled': <bool>}
+        """
+        courserun_key = CourseKey.from_string(course_id)
+
+        data = {
+            'enabled': learning_assistant_enabled(courserun_key),
```
MichaelRoytman, line 128: Note: This will look at the `CourseWaffleFlag` and the `LearningAssistantCourseEnabled` model.

learning_assistant/views.py
```diff
@@ -91,3 +91,41 @@ def post(self, request, course_id):
         status_code, message = get_chat_response(prompt_template, message_list, course_id)
 
         return Response(status=status_code, data=message)
+
+
+class LearningAssistantEnabledView(APIView):
+    """
+    View to retrieve whether the Learning Assistant is enabled for a course.
```
varshamenon4, line 98: Nice docs!

learning_assistant/views.py
```diff
@@ -91,3 +91,41 @@ def post(self, request, course_id):
...
+
+    Accepts: [GET]
+
+    Path: /learning_assistant/v1/course_id/{course_key}/enabled
+
+    Parameters:
+        * course_key: the ID of the course
+
+    Responses:
+        * 200: OK
```
varshamenon4, line 111: I'm guessing that since this is a GET which checks for enabled, we don't need to do any validation? For example, would we need to validate and return an 4** response if the course id doesn't exist?

learning_assistant/views.py
```diff
@@ -91,3 +91,41 @@ def post(self, request, course_id):
...
+
+    def get(self, request, course_id):
+        """
+        Given a course ID, retrieve whether the Learning Assistant is enabled for the corresponding course.
+
+        The response will be in the following format.
+
+            {'enabled': <bool>}
+        """
+        courserun_key = CourseKey.from_string(course_id)
```
varshamenon4, line 125: [nit] is there a reason to use `courserun_key` rather than course_key here? I like it if it is more specific but I wasn't sure if these are just the same? If so, I might suggest renaming to simplify.

tests/test_views.py
```diff
@@ -164,3 +165,31 @@ def test_chat_response(self, mock_mode, mock_enrollment, mock_role, mock_waffle,
...
+        super().setUp()
+        self.course_id = 'course-v1:edx+test+23'
+
+    @ddt.data(
+        (True, True),
+        (False, False),
+    )
+    @ddt.unpack
+    @patch('learning_assistant.views.learning_assistant_enabled')
+    def test_learning_assistant_enabled(self, mock_value, expected_value, mock_learning_assistant_enabled):
```
varshamenon4, line 190: Nice testing!

tests/test_views.py
```diff
@@ -164,3 +165,31 @@ def test_chat_response(self, mock_mode, mock_enrollment, mock_role, mock_waffle,
...
+    sys.modules['lms.djangoapps.courseware.toggles'] = MagicMock()
+    sys.modules['common.djangoapps.course_modes.models'] = MagicMock()
+    sys.modules['common.djangoapps.student.models'] = MagicMock()
+
+    def setUp(self):
+        super().setUp()
+        self.course_id = 'course-v1:edx+test+23'
+
+    @ddt.data(
+        (True, True),
```
varshamenon4, line 185: [nit] since these are the same (in other words, if the mock value is true, then the expected should also be true) would you need to have both? Could you reduce to one so that there are fewer params?

varshamenon4: I could see the value of having both for readability though, so would defer to you!

varshamenon4 (commented):
Looks great! Just a couple of comments and nits.

learning_assistant/views.py
MichaelRoytman: That's a good question. I don't think there's a spot in the code path where the existence of the course associated with the course ID is checked. [learning_assistant_enabled](https://github.com/edx/learning-assistant/blob/main/learning_assistant/api.py#L120) queries the `LearningAssistantCourseEnabled` model and handles the case where there is no instance for a given course ID. It also queries the course waffle flag, which shouldn't error if the course ID doesn't exist; it would just return `False` in that case.

I do think it would make sense to add a `try...except` around the `CourseKey.from_string(course_id)` line, in case the course ID is malformatted and can't be parsed into a `CourseKey`. I think the course pattern regular expression in the URL should take care of validating the course ID, but it doesn't hurt to be extra cautious here.

What do you think?

learning_assistant/views.py
MichaelRoytman: I copied some of this code from the view about, which Alie wrote, but I think there is a reason. An instance of a `CourseKey` technically represents a run of a course, or a course run. There is a difference between a canonical or "shell" course and the actual instance of the course, which is a course run. I think, in most places, we just use `course_key` to refer to `CourseKey` or the course run key, because it's simpler and an implementation detail. The reason we're being a little more explicit here is that we actually do want the course key in some places.

The Xpert index that we use to pull data from on the Xpert Platform backend expects a course key. That is, they do not want the course run key. So we're handling course run keys and course keys in this code base, which is why I think Alie chose a more specific name. You can see where we "convert" a course run key into a course key [here](https://github.com/edx/learning-assistant/blob/8a25061b1afc855f1f578965797bb7c4864eb08e/learning_assistant/utils.py#L63).

If the use of `courserun_key` is confusing here, I could change it to `course_key`. I don't think the distinction in this view is super important.

tests/test_views.py
MichaelRoytman: It feels a little more explicit to me, personally. This way, I'm saying "this is the input, and this is the expected output". In my mind, if I set the ddt data to be `(True, False)` (i.e. "this is the input"), it requires reading the test in order to understand what the expected case is. That's not a huge deal, but I prefer to be explicit, when I can.

learning_assistant/views.py
varshamenon4: That makes sense! Thanks for clarifying! I'm good with keeping it as is, as like you said, it's a little more explicit. Resolving now.

tests/test_views.py
varshamenon4: I think that's a good decision to be more explicit!

varshamenon4 (approved):
lgtm! (I think that you might jut need a test for the invalid key error you've just added, but otherwise looks great!)

PR #67, opened 2024-02-13
feat: replace waffle flag with setting

alangsto:
## [COSMO-191](https://2u-internal.atlassian.net/browse/COSMO-191)

To prepare for a wider release, the use of a course waffle flag to enable the learning assistant should be replaced by a platform wide django setting. 

If no setting is defined, the feature will default to a disabled state.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -6,7 +6,7 @@ from unittest.mock import MagicMock, patch
 
 import ddt
 from django.core.cache import cache
-from django.test import TestCase
+from django.test import TestCase, override_settings
 from opaque_keys.edx.keys import CourseKey, UsageKey
 
 from learning_assistant.api import (
@@ -14,6 +14,7 @@ from learning_assistant.api import (
     _get_children_contents,
     _leaf_filter,
     get_block_content,
+    learning_assistant_available,
     learning_assistant_enabled,
     render_prompt_template,
     set_learning_assistant_enabled,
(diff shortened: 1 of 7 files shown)


tests/test_api.py
```diff
@@ -206,7 +207,7 @@ def test_render_prompt_template(self, unit_content, flag_enabled, mock_get_conte
 @ddt.ddt
 class LearningAssistantCourseEnabledApiTests(TestCase):
     """
-    Test suite for the learning_assistant_enabled and set_learning_assistant_enalbed api functions.
+    Test suite for learning_assistant_available, learning_assistant_enabled, and set_learning_assistant_enalbed.
```
MichaelRoytman, line 210: Nit.
```suggestion
    Test suite for learning_assistant_available, learning_assistant_enabled, and set_learning_assistant_enabled.
```

tests/test_api.py
```diff
@@ -223,7 +224,8 @@ def setUp(self):
         (False, False, False, False),
     )
     @ddt.unpack
-    @patch('learning_assistant.api.learning_assistant_available')
+    @patch('learning_assistant.api.learning_assistant_available_flag')
+    @override_settings(LEARNING_ASSISTANT_AVAILABLE=False)
```
MichaelRoytman, line 228: Do these tests fail if `LEARNING_ASSISTANT_AVAILABLE=True`? Or are you just trying to ensure that these tests are testing the waffle flag logic and not the Django setting?

MichaelRoytman (approved):

tests/test_api.py
alangsto: I was trying to limit the scope of this test, and not introduce more booleans to test the logic of the django setting. Although now I'm realizing because I added a test for the api function `learning_assistant_available`, I could mock that here, and avoid overriding the setting.

PR #70, opened 2024-02-20
feat: remove use of course waffle flag

alangsto:
## [COSMO-191](https://2u-internal.atlassian.net/browse/COSMO-191)

Remove all uses of the course waffle flag now that the django setting related code has been deployed, and the setting has been enabled.


varshamenon4 (approved):
Awesome!

PR #72, opened 2024-02-26
feat: use course cache to retrieve course data

alangsto:
## [COSMO-22](https://2u-internal.atlassian.net/browse/COSMO-202)

The Xpert Platform course index does not include courses based on certain enrollment dates and other discovery fields, like `is_marketable`. This means that Xpert is only available to a more limited set of courses. 

To increase the number of courses in which Xpert is available, we can opt to not rely on the Xpert Platform backend to inject the course title and course skills into the prompt template, and instead use an [already-defined cache](https://github.com/openedx/edx-platform/blob/c61df904c1d2a5f523f1da44460c21e17ec087ee/openedx/core/djangoapps/catalog/utils.py#L767C5-L767C20) in the LMS to retrieve that same data.


learning_assistant/views.py
```diff
@@ -93,7 +93,12 @@ def post(self, request, course_id):
             }
         )
 
-        prompt_template = render_prompt_template(request, request.user.id, course_id, unit_id)
+        # the "course id" being passed in as a query parameter is really a course run id. For the sake of
+        # clarity here, the value is reassigned to a more aptly named variable.
+        course_run_id = course_id
```
alangsto, line 98: The current variable name is a little confusing. The query parameter should really be `course_run_id`, but that would require a change to the URL name and an update to the frontend to account for that. 

This change that I've added here should be temporary, as I needed to move the call to `get_course_id` into the view, as both `render_prompt_template` and `get_chat_response` need the course id.

docs/decisions/0004-course-cache-use.rst
```diff
@@ -0,0 +1,37 @@
...
+**Accepted** *2024-02-26*
+
+Context
+*******
+Each course run ID in edx-platform is associated with another course ID. While for many courses, the mapping between
+course run ID and course ID is straight forward, i.e. edX+testX+2019 -> edX+testX, this is not the case for every
+course on edX. The discovery service is the source of truth for mappings between course run ID and course IDs, and
+string manipulation cannot be relied on as an accurate way to map between the two forms of ID.
+
+The learning-assistant CourseChatView NEED LINK accepts a course run ID as a query parameter, but parts of the python API defined
```
MichaelRoytman, line 16: Do you need to add a link here?

docs/decisions/0004-course-cache-use.rst
```diff
@@ -0,0 +1,37 @@
...
+in the learning assistant backend also require the course ID associated with a given course.
+
+Decision
+********
+In order to determine the mapping between a course run ID and course ID in the learning-assistant app, we will make
```
MichaelRoytman, line 21: I think we should also mention the course cache to get the skill names and title.

tests/test_api.py
```diff
@@ -183,20 +183,25 @@ def test_get_block_content(self, mock_get_children_contents, mock_get_single_blo
...
+    @patch('learning_assistant.api.get_cache_course_data')
     @patch('learning_assistant.toggles._is_learning_assistant_waffle_flag_enabled')
     @patch('learning_assistant.api.get_block_content')
-    def test_render_prompt_template(self, unit_content, flag_enabled, mock_get_content, mock_is_flag_enabled):
+    def test_render_prompt_template(
+        self, unit_content, flag_enabled, mock_get_content, mock_is_flag_enabled, mock_cache
+    ):
         mock_get_content.return_value = (len(unit_content), unit_content)
         mock_is_flag_enabled.return_value = flag_enabled
+        mock_cache.return_value = {'skill_names': ['skills'], 'title': ['title']}
```
MichaelRoytman, line 194: I think we should assert that the title and skills appear in `prompt_text` somewhere. Do you think this is a good place for it?

docs/decisions/0004-course-cache-use.rst
alangsto: Yes, thanks for catching that

tests/test_api.py
alangsto: Yes, I can add that check!

docs/decisions/0004-course-cache-use.rst
```diff
@@ -0,0 +1,49 @@
...
+being used to inject course skill names and course titles into the system prompt (see `System Prompt Design Changes`_ for
+original details), were too limited. Courses included in that index were conditionally added depending on course
+enrollment dates and additional fields from the discovery course API. While the 2U Xpert Platform team may work to address
+the gap in product needs for their current course index, an alternate method for retrieving course skills and title should
+be considered.
+
+Decision
+********
+In order to determine the mapping between a course run ID and course ID in the learning-assistant app, we will make
+use of an `existing course run cache that is defined in edx-platform`_. Similarly, to retrieve the skill names and title of a course, we will also use
```
MichaelRoytman, line 29: This link isn't rendering properly because the text here and in the link definition are different (this one has "run" and the one below doesn't).

docs/decisions/0004-course-cache-use.rst
```diff
@@ -0,0 +1,49 @@
...
+an `existing course cache`_. Both caches store data from the discovery API for course runs and courses, respectively.
```
MichaelRoytman, line 30: It looks like this link is missing a definition below.

MichaelRoytman (approved):
This looks good! Two of the links in the ADR are not formatted correctly, so it would be great if you could fix those before merging. Thanks!

PR #95, opened 2024-06-27
feat: support variation param for model and prompt

Zacharis278:
Adds an optional `response_variation` parameter when posting to the course chat endpoint to use a different model and/or prompt.

### JIRA [COSMO-256](https://2u-internal.atlassian.net/browse/COSMO-356)

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -105,7 +105,7 @@ def get_block_content(request, user_id, course_id, unit_usage_key):
 
 def render_prompt_template(request, user_id, course_run_id, unit_usage_key, course_id, template_string):
     """
-    Return a rendered prompt template, specified by the LEARNING_ASSISTANT_PROMPT_TEMPLATE setting.
+    Return a rendered prompt template.
     """
     unit_content = ''
 
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -103,7 +103,7 @@ def get_block_content(request, user_id, course_id, unit_usage_key):
     return cache_data['content_length'], cache_data['content_items']
 
 
-def render_prompt_template(request, user_id, course_run_id, unit_usage_key, course_id):
+def render_prompt_template(request, user_id, course_run_id, unit_usage_key, course_id, template_string):
     """
     Return a rendered prompt template, specified by the LEARNING_ASSISTANT_PROMPT_TEMPLATE setting.
     """
@@ -117,7 +117,6 @@ def render_prompt_template(request, user_id, course_run_id, unit_usage_key, cour
     skill_names = course_data['skill_names']
     title = course_data['title']
 
-    template_string = getattr(settings, 'LEARNING_ASSISTANT_PROMPT_TEMPLATE', '')
     template = Environment(loader=BaseLoader).from_string(template_string)
     data = template.render(unit_content=unit_content, skill_names=skill_names, title=title)
     return data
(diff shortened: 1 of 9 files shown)


learning_assistant/views.py
```diff
@@ -95,9 +98,17 @@ def post(self, request, course_run_id):
 
         course_id = get_course_id(course_run_id)
 
-        prompt_template = render_prompt_template(request, request.user.id, course_run_id, unit_id, course_id)
+        if response_variation == ResponseVariations.GPT4_UPDATED_PROMPT:
+            gpt_model = GptModels.GPT_4o
+            template_string = getattr(settings, 'LEARNING_ASSISTANT_EXPERIMENTAL_PROMPT_TEMPLATE', '')
+        else:
+            gpt_model = GptModels.GPT_3_5_TURBO
```
Zacharis278, line 105: TODO: Make sure this is _actually_ the default before merging

learning_assistant/api.py
```diff
@@ -103,7 +103,7 @@ def get_block_content(request, user_id, course_id, unit_usage_key):
     return cache_data['content_length'], cache_data['content_items']
 
 
-def render_prompt_template(request, user_id, course_run_id, unit_usage_key, course_id):
+def render_prompt_template(request, user_id, course_run_id, unit_usage_key, course_id, template_string):
```
varshamenon4, line 106: Questioin: is the template_string required? Either way, would we need to handle error cases?

tests/test_views.py
```diff
@@ -155,13 +157,15 @@ def test_invalid_messages(self, mock_role, mock_waffle, mock_render):
     @patch('learning_assistant.views.get_user_role')
     @patch('learning_assistant.views.CourseEnrollment.get_enrollment')
     @patch('learning_assistant.views.CourseMode')
-    def test_chat_response(self, mock_mode, mock_enrollment, mock_role, mock_waffle, mock_chat_response, mock_render):
+    def test_chat_response_default(
```
varshamenon4, line 160: Question: are there any error cases/non-happy paths that should be tested?

varshamenon4 (commented):
Looks good! Just a couple of questions.

learning_assistant/api.py
```diff
@@ -103,7 +103,7 @@ def get_block_content(request, user_id, course_id, unit_usage_key):
...
     """
     Return a rendered prompt template, specified by the LEARNING_ASSISTANT_PROMPT_TEMPLATE setting.
```
rijuma, line 108: If `LEARNING_ASSISTANT_PROMPT_TEMPLATE` is remove, we might want to update this description.

learning_assistant/api.py
Zacharis278: yeah it has no default so you'd need to pass this. IMO I don't think we need to null check each param here, the linter already covers us since won't let you invoke a function w/ missing arguments

learning_assistant/api.py
Zacharis278: good catch 👍

tests/test_views.py
Zacharis278: humm yeah the only input we're adding is the query param, bad data for everything else is already tested above. I think I should validate an invalid variation doesn't do anything unexpected tho.

Zacharis278: okay, added a case for an invalid variation and noticed we needed better coverage on what template gets loaded since the way this was mocked left some holes

learning_assistant/views.py
Zacharis278: updated based on the default here: https://github.com/2uinc/ai-chatbot-backend/blob/7fbc9498018ad3228a17ba4e8567ebc4320baf7a/xpert/utils/openai.py#L47C54-L47C72

rijuma (approved):
LGTM

PR #110, opened 2024-09-19
fix: handle invalid unit ids

alangsto:
## [COSMO-432](https://2u-internal.atlassian.net/browse/COSMO-432)

Occasionally requests coming from the frontend contain a `null` unit ID. This causes an error, as the null value is interpreted as a string, and then used as a unit ID although it is invalid. Via local testing, I also found that `last` was a unit ID value that could be passed along in the POST request to our Xpert backend. 

To avoid unhandled 500 responses, we should catch the `InvalidKeyError` being thrown, and log when it is happening. The request should still complete successfully, even without a valid unit ID, as content is not necessary for making a successful request.

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from django.core.cache import cache
 from edx_django_utils.cache import get_cache_key
 from jinja2 import BaseLoader, Environment
-from opaque_keys.edx.keys import CourseKey
+from opaque_keys import InvalidKeyError
 
 from learning_assistant.constants import ACCEPTED_CATEGORY_TYPES, CATEGORY_TYPE_MAP
 from learning_assistant.data import LearningAssistantCourseEnabledData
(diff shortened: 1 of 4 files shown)


learning_assistant/api.py
```diff
@@ -109,9 +108,15 @@ def render_prompt_template(request, user_id, course_run_id, unit_usage_key, cour
...
     unit_content = ''
 
-    course_run_key = CourseKey.from_string(course_run_id)
-    if unit_usage_key and course_content_enabled(course_run_key):
-        _, unit_content = get_block_content(request, user_id, course_run_id, unit_usage_key)
+    if unit_usage_key:
+        try:
+            _, unit_content = get_block_content(request, user_id, course_run_id, unit_usage_key)
+        except InvalidKeyError:
+            log.info(
```
schenedx, line 115: Are you sure we want to log `info` here?
Why not `error` or `warning`?

alangsto: Warning might be more appropriate. I wasn't sure about logging an error because I wasn't sure if it would impact any DD monitors watching error rates in logs (if those exist).

The reason I initially chose info was because it's something for us to keep track of to better understand the scope of this and when it occurs, but it's not an indication that the request won't succeed.

Zacharis278 (approved):

schenedx (approved):
👍

PR #114, opened 2024-10-15
feat: use default prompt template

alangsto:
## [COSMO-475](https://2u-internal.atlassian.net/browse/COSMO-475)

As part of the experiment clean up, the experimental prompt has been adopted as the "default" prompt. The backend should be updated to now use the default/non-experimental setting.


schenedx (approved):
👍

PR #119, opened 2024-10-25
feat: GET endpoint to retrieve message history

ilee2u:
Ticket: https://2u-internal.atlassian.net/browse/COSMO-434

Add GET endpoint to learning-assistant to retrieve message history for a user/course combination.
- Should get messages by user/course combination
- Should add query parameter to limit the amount of messages we receive

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -189,11 +189,12 @@ def get_course_id(course_run_id):
     return course_key
 
 
-def get_message_history(course_id, user):
+def get_message_history(course_id, user, message_limit):
     """
     Given ARG, return RET
     """
-    # NOTE: How much should we limit the message history to?
+    # NOTE: message_limit = how many messages back we go
     # note for self: this is descending order for the created date, which shows the latest first.
-    message_history = LearningAssistantMessage.objects.filter(course_id=course_id, user=user).order_by('-created')[:50]
+    message_history = LearningAssistantMessage.objects.filter(
+        course_id=course_id, user=user).order_by('-created')[:message_limit]
     return message_history.content
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -195,13 +195,6 @@ def get_message_history(course_id, user, message_count):
 
     Returns a number of messages equal to the message_count value.
     """
-    # If the received message count exceeds the number of messages present, we return all the messages queried.
-    actual_message_count = LearningAssistantMessage.objects.count()
-    if message_count > actual_message_count:
-        amount_to_get = actual_message_count
-    else:
-        amount_to_get = message_count
-
     message_history = LearningAssistantMessage.objects.filter(
-        course_id=course_id, user=user).order_by('-created')[:amount_to_get]
+        course_id=course_id, user=user).order_by('-created')[:message_count]
     return message_history
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -189,12 +189,12 @@ def get_course_id(course_run_id):
     return course_key
 
 
-def get_message_history(course_id, user, message_limit):
+def get_message_history(course_id, user, message_count):
     """
     Given ARG, return RET
     """
-    # NOTE: message_limit = how many messages back we go
+    # NOTE: message_count = how many messages back we go
     # note for self: this is descending order for the created date, which shows the latest first.
     message_history = LearningAssistantMessage.objects.filter(
-        course_id=course_id, user=user).order_by('-created')[:message_limit]
+        course_id=course_id, user=user).order_by('-created')[:message_count]
     return message_history.content
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -195,7 +195,6 @@ def get_message_history(course_id, user, message_count):
 
     Returns a number of messages equal to the message_count value.
     """
-
     # If the received message count exceeds the number of messages present, we return all the messages queried.
     actual_message_count = LearningAssistantMessage.objects.count()
     if message_count > actual_message_count:
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -195,6 +195,14 @@ def get_message_history(course_id, user, message_count):
 
     Returns a number of messages equal to the message_count value.
     """
+
+    # If the received message count exceeds the number of messages present, we return all the messages queried.
+    actual_message_count = LearningAssistantMessage.objects.count()
+    if message_count > actual_message_count:
+        amount_to_get = actual_message_count
+    else:
+        amount_to_get = message_count
+
     message_history = LearningAssistantMessage.objects.filter(
-        course_id=course_id, user=user).order_by('-created')[:message_count]
-    return message_history.content
+        course_id=course_id, user=user).order_by('-created')[:amount_to_get]
+    return message_history
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -11,7 +11,7 @@ from opaque_keys import InvalidKeyError
 
 from learning_assistant.constants import ACCEPTED_CATEGORY_TYPES, CATEGORY_TYPE_MAP
 from learning_assistant.data import LearningAssistantCourseEnabledData
-from learning_assistant.models import LearningAssistantCourseEnabled
+from learning_assistant.models import LearningAssistantCourseEnabled, LearningAssistantMessage
 from learning_assistant.platform_imports import (
     block_get_children,
     block_leaf_filter,
@@ -187,3 +187,13 @@ def get_course_id(course_run_id):
     course_data = get_cache_course_run_data(course_run_id, ['course'])
     course_key = course_data['course']
     return course_key
+
+
+def get_message_history(course_id, user):
+    """
+    Given ARG, return RET
+    """
+    # NOTE: How much should we limit the message history to?
+    # note for self: this is descending order for the created date, which shows the latest first.
+    message_history = LearningAssistantMessage.objects.filter(course_id=course_id, user=user).order_by('-created')[:50]
+    return message_history.content
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -191,10 +191,10 @@ def get_course_id(course_run_id):
 
 def get_message_history(course_id, user, message_count):
     """
-    Given ARG, return RET
+    Given a course run id (str), user (User), and message count (int), return the associated message history.
+
+    Returns a number of messages equal to the message_count value.
     """
-    # NOTE: message_count = how many messages back we go
-    # note for self: this is descending order for the created date, which shows the latest first.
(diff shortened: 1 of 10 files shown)


learning_assistant/views.py
```diff
@@ -149,3 +154,69 @@ def get(self, request, course_run_id):
...
+    authentication_classes = (SessionAuthentication, JwtAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request, course_run_id):
+        """
+        Given a course run ID, retrieve the message history for the corresponding user.
+
+        The response will be in the following format.
+
+            {'enabled': <bool>}
```
alangsto, line 187: NIT: this isn't the correct response for this view

learning_assistant/views.py
```diff
@@ -149,3 +154,69 @@ def get(self, request, course_run_id):
...
+        """
+
```
alangsto, line 189: We should have a query parameter as part of this request to denote how many messages we want to retrieve. That could then be passed to your python api function to limit how many messages we retrieve

learning_assistant/urls.py
```diff
@@ -19,4 +23,9 @@
         LearningAssistantEnabledView.as_view(),
         name='enabled',
     ),
+    re_path(
+        fr'learning_assistant/v1/course_id/{COURSE_ID_PATTERN}/history',
+        LearningAssistantMessageHistoryView.as_view(),
+        name='enabled',
```
alangsto, line 29: the name should be updated

learning_assistant/views.py
```diff
@@ -149,3 +154,69 @@ def get(self, request, course_run_id):
...
+                status=http_status.HTTP_403_FORBIDDEN,
+                data={'detail': 'Learning assistant not enabled for course.'}
+            )
+
+        # If user does not have an enrollment record, or is not staff, they should not have access
+        user_role = get_user_role(request.user, courserun_key)
+        enrollment_object = CourseEnrollment.get_enrollment(request.user, courserun_key)
+        enrollment_mode = enrollment_object.mode if enrollment_object else None
+        if (
+            (enrollment_mode not in CourseMode.VERIFIED_MODES)
```
schenedx, line 210: Are you sure this logic condition is correct?
We should return if user_role_is_staff, no?
Are you sure we don't want to show message history to audit mode learners?

learning_assistant/api.py
```diff
@@ -187,3 +187,13 @@ def get_course_id(course_run_id):
...
     return course_key
+
+
+def get_message_history(course_id, user):
+    """
+    Given ARG, return RET
+    """
+    # NOTE: How much should we limit the message history to?
+    # note for self: this is descending order for the created date, which shows the latest first.
+    message_history = LearningAssistantMessage.objects.filter(course_id=course_id, user=user).order_by('-created')[:50]
```
schenedx, line 198: Only the last 50 records? Should that be a config value?

learning_assistant/views.py
ilee2u: Not quite sure how to format this here: https://github.com/edx/learning-assistant/pull/119/files#diff-bf828b0d52eaac58c68450b8851259f392b937ae1a26b7cc74749fe9afe970beR183

learning_assistant/views.py
ilee2u: Not quite sure, I copy/pasted this from another view so this could be wrong. @alangsto Do you have an opinion on this?

learning_assistant/api.py
```diff
@@ -187,3 +187,14 @@ def get_course_id(course_run_id):
     course_data = get_cache_course_run_data(course_run_id, ['course'])
     course_key = course_data['course']
     return course_key
+
+
+def get_message_history(course_id, user, message_count):
+    """
+    Given a course run id (str), user (User), and message count (int), return the associated message history.
+
+    Returns a number of messages equal to the message_count value.
```
varshamenon4, line 196: Nit: If the message count exceeds the number of messages, we could return messages that are less than this value (in other words, the number of messages cannot exceed this value but could be less than this value).

learning_assistant/api.py
```diff
@@ -187,3 +187,14 @@ def get_course_id(course_run_id):
...
+    """
+    message_history = LearningAssistantMessage.objects.filter(
+        course_id=course_id, user=user).order_by('-created')[:message_count]
```
varshamenon4, line 199: Would we need the handle the case here if the message count exceeds the number of messages? (i.e. to prevent and index error)

learning_assistant/views.py
```diff
@@ -149,3 +154,75 @@ def get(self, request, course_run_id):
...
+            )
+
+        if not learning_assistant_enabled(courserun_key):
+            return Response(
+                status=http_status.HTTP_403_FORBIDDEN,
+                data={'detail': 'Learning assistant not enabled for course.'}
+            )
+
+        # If user does not have an enrollment record, or is not staff, they should not have access
+        # NOTE: Do we need this? Do we currently not have any plans to show message history to audit mode learners?
```
varshamenon4, line 211: I think we do want to have this logic for now and then will adjust once we add the audit experience work. We'll need to have logic to limit access to audit learners.

varshamenon4 (commented):
Looks good so far! A couple of questions/comments!

learning_assistant/views.py
alangsto: It should be a list of messages, so something like [{'role': 'assistant', 'content': 'something'}]

learning_assistant/views.py
```diff
@@ -149,3 +154,75 @@ def get(self, request, course_run_id):
...
+
+    Responses:
+        * 200: OK
+        * 400: Malformed Request - Course ID is not a valid course ID.
+    """
+
+    authentication_classes = (SessionAuthentication, JwtAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request, course_run_id, message_count):
```
alangsto, line 181: I think the message count should be a query parameter, not a path parameter. If no message count is provided as a query parameter, we should have a default value.

learning_assistant/views.py
```diff
@@ -149,3 +154,75 @@ def get(self, request, course_run_id):
...
+            and not user_role_is_staff(user_role)
+        ):
+            return Response(
+                status=http_status.HTTP_403_FORBIDDEN,
+                data={'detail': 'Must be staff or have valid enrollment.'}
+            )
+
+        course_id = get_course_id(course_run_id)
+        user = request.user
+        data = get_message_history(course_id, user, message_count)
```
alangsto, line 226: Before being returned, this should be serialized (like https://github.com/edx/edx-exams/blob/d8cf60a98bd86be4cab31f16ae5969a7c52ef396/edx_exams/apps/api/v1/views.py#L227). This gives us better control over what we return.

learning_assistant/api.py
```diff
@@ -187,3 +187,21 @@ def get_course_id(course_run_id):
...
+
+
+def get_message_history(course_id, user, message_count):
+    """
+    Given a course run id (str), user (User), and message count (int), return the associated message history.
+
+    Returns a number of messages equal to the message_count value.
+    """
+    # If the received message count exceeds the number of messages present, we return all the messages queried.
+    actual_message_count = LearningAssistantMessage.objects.count()
```
alangsto, line 199: I actually don't know if this is strictly necessary. Have you tested the code without this, and just indexing using the `message_count` value? Only asking because the same syntax is used here: https://github.com/openedx/edx-platform/blob/b20498cc4eec4ed1b85508967bef6c915372dd5b/openedx/core/djangoapps/notifications/tasks.py#L105

learning_assistant/views.py
```diff
@@ -149,3 +154,69 @@ def get(self, request, course_run_id):
...
+
+    Responses:
+        * 200: OK
+        * 400: Malformed Request - Course ID is not a valid course ID.
+    """
+
+    authentication_classes = (SessionAuthentication, JwtAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def get(self, request, course_run_id, message_count=50):
```
alangsto, line 181: I still think this should be a query parameter, not a path parameter.

tests/test_views.py
```diff
@@ -210,16 +213,129 @@ def setUp(self):
...
+        )
+        self.assertEqual(response.status_code, 403)
+
+    @patch('learning_assistant.views.learning_assistant_enabled')
+    @patch('learning_assistant.views.get_user_role')
+    @patch('learning_assistant.views.CourseEnrollment.get_enrollment')
+    @patch('learning_assistant.views.CourseMode')
+    @patch('learning_assistant.views.get_course_id')
+    def test_learning_message_history_view_get(self,
+                                               mock_get_course_id,
```
alangsto, line 309: NIT: This spacing looks a little strange to me, can you remove the indents?

learning_assistant/urls.py
```diff
@@ -19,4 +19,9 @@
         LearningAssistantEnabledView.as_view(),
         name='enabled',
     ),
+    re_path(
+        fr'learning_assistant/v1/course_id/{COURSE_ID_PATTERN}/history/<int:message_count>',
```
alangsto, line 23: the message count shouldn't be specified in the path, in my opinion. Happy to be convinced otherwise, but it feels more appropriate, as the message count param is used to filter on the resource being specified.

learning_assistant/views.py
ilee2u: I keep mixing these up- I think I've got it right now, gonna use `request.GET.get('message_count', 50) instead.

Side note, what do you think the default value should be? Is 50 good?

tests/test_views.py
ilee2u: this was throwing a "line too long" quality error, I might just ignore it instead.

learning_assistant/urls.py
ilee2u: Just realizing query parameters don't need to be in the URL... I'll get rid of that now

tests/test_views.py
ilee2u: Actually, I a better way by just moving the "self" param to the next line, then moving the spacing of all the params back to one tab past the function declaration.

learning_assistant/api.py
ilee2u: Oh that actually works. I'll get rid of this code.

learning_assistant/views.py
alangsto: Yes, that should be the right syntax! And 50 seems good.

tests/test_views.py
alangsto: You can also keep them on the same line, they don't each need to be on their own!

schenedx (approved):
👍

alangsto (approved):
A few nits but otherwise looks good!

alangsto:
Forgot one other NIT, but could you add an entry to the changelog for this? Thank you!

PR #121, opened 2024-10-30
feat: Adding chat messages to the DB

rijuma:
This update adds the last user message to the course chat view.

I've also added some constants to `LearningAssistantMessage` model for consistency.

> [!WARNING]
> I've added an extra validation on the endpoint requiring the last message from the chat to be from the user.


learning_assistant/views.py
```diff
@@ -76,6 +79,15 @@ def post(self, request, course_run_id):
         unit_id = request.query_params.get('unit_id')
 
         message_list = request.data
+
+        # Check that the last message in the list corresponds to a user
+        new_user_message = message_list[-1]
```
schenedx, line 84: I don't really understand the code block and the comment. Can you explain more to help someone like to understand the logic better?

tests/test_views.py
```diff
@@ -170,16 +172,29 @@ def test_chat_response_default(
 
         test_data = [
             {'role': 'user', 'content': 'What is 2+2?'},
-            {'role': 'assistant', 'content': 'It is 4'}
+            {'role': 'assistant', 'content': 'It is 4'},
+            {'role': 'user', 'content': 'And what else?'},
```
rijuma, line 176: I've added a check that validates the last message from the list must be from a user. Otherwise its confusing to store the values. Also, I think we should be processing only one question at the time.

learning_assistant/views.py
schenedx: What if the message_list array is empty?

learning_assistant/models.py
```diff
@@ -35,7 +35,15 @@ class LearningAssistantMessage(TimeStampedModel):
...
+
+    Roles = (
+        (USER_ROLE, USER_ROLE),
+        (ASSISTANT_ROLE, ASSISTANT_ROLE),
+    )
+
     course_id = CourseKeyField(max_length=255, db_index=True)
     user = models.ForeignKey(USER_MODEL, db_index=True, on_delete=models.CASCADE)
-    role = models.CharField(max_length=64)
+    role = models.CharField(choices=Roles, max_length=64)
```
rijuma, line 48: I think this doesn't affect the database, so no migration needed.

learning_assistant/views.py
rijuma: I assumed the endpoint was to process a new user message. Maybe I misunderstood it but if it is, then it should include at least one user message.

learning_assistant/views.py
```diff
@@ -35,6 +38,27 @@ class CourseChatView(APIView):
     authentication_classes = (SessionAuthentication, JwtAuthentication,)
     permission_classes = (IsAuthenticated,)
 
+    def __save_user_interaction(self, user_id, user_message, assistant_message):
```
alangsto, line 41: can this be moved to the `api.py` file? I also think it might make more sense for it to look like `save_user_message(user_id, role, message)`, and call it twice from within the view.

learning_assistant/views.py
```diff
@@ -103,6 +136,12 @@ def post(self, request, course_run_id):
         )
         status_code, message = get_chat_response(prompt_template, message_list)
 
+        self.__save_user_interaction(
```
alangsto, line 139: This needs to be gated by a waffle flag, which also needs to be created.

rijuma: Oh shoot. I totally forgot about it.

learning_assistant/views.py
rijuma: Just to point out: We discussed this offline and came to the conclusion that the approach is correct.

learning_assistant/views.py
rijuma: Moved to the API module and added two calls in the view.

learning_assistant/views.py
rijuma: Added waffle flag and gated the functionality behind it.

tests/test_views.py
```diff
@@ -14,10 +14,12 @@
 from django.test.client import Client
 from django.urls import reverse
 
+from learning_assistant.models import LearningAssistantMessage
+
 User = get_user_model()
 
 
-class TestClient(Client):
+class FakeClient(Client):
```
rijuma, line 22: This class was renamed to remove the following test warning:
```
tests/test_views.py:22
  /Users/(ueser)/edx/src/learning-assistant/tests/test_views.py:22: PytestCollectionWarning: cannot collect test class 'TestClient' because it has a __init__ constructor (from: tests/test_views.py)
    class TestClient(Client):
```

learning_assistant/views.py
```diff
@@ -76,6 +78,21 @@ def post(self, request, course_run_id):
...
+        if new_user_message['role'] != LearningAssistantMessage.USER_ROLE:
+            return Response(
+                status=http_status.HTTP_400_BAD_REQUEST,
+                data={'detail': "Expects user role on last message."}
+            )
+
+        course_id = get_course_id(course_run_id)
+        user_id = request.user.id
+
+        if chat_history_enabled(course_id):
```
alangsto, line 93: Should this be the course_id or a course key?

rijuma: Good catch. According to the code to check the flag we [check for Learning Assistant enabled state](https://github.com/edx/learning-assistant/blob/e28918a6a682719c72533f0435dcd76488ff0e1b/learning_assistant/views.py#L158) is a CourseKey based on the course run. I'll update it quickly.

learning_assistant/models.py
schenedx: Are you sure? Did you run `makemigrations`?

schenedx (approved):
👍

learning_assistant/api.py
```diff
@@ -189,6 +191,24 @@ def get_course_id(course_run_id):
     return course_key
 
 
+def save_chat_message(user_id, chat_role, message):
```
alangsto, line 194: This function should also take in a course_id, and use that as a field in the message being saved

alangsto (changes requested):

learning_assistant/models.py
rijuma: I didn't. I'm sorry for that.
Just added the migration file.

learning_assistant/api.py
rijuma: Nice catch. I wonder why this wasn't detected on the unit tests, maybe there's a constraint missing?

rijuma: Just updated `save_chat_message()` to take the Courserun Key and save it to the model.

rijuma:
Just to confirm, I've validated locally that no message is saved without enabling the feature flag and this is the result after enabling it:
```
+----+----------------------------+----------------------------+---------------------------------+-----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------+
| id | created                    | modified                   | course_id                       | role      | content                                                                                                                                                              | user_id |
+----+----------------------------+----------------------------+---------------------------------+-----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------+
|  1 | 2024-11-04 19:05:07.403363 | 2024-11-04 19:05:07.403363 | course-v1:edX+DemoX+Demo_Course | user      | marco                                                                                                                                                                |       3 |
|  2 | 2024-11-04 19:05:21.357636 | 2024-11-04 19:05:21.357636 | course-v1:edX+DemoX+Demo_Course | assistant | It seems like you might be looking for something specific. Could you please clarify your question or let me know how I can assist you with the Demonstration Course? |       3 |
+----+----------------------------+----------------------------+---------------------------------+-----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------+
```

alangsto (approved):

PR #122, opened 2024-10-31
feat: add management command to remove expired messages

alangsto:
## [COSMO-554](https://2u-internal.atlassian.net/browse/COSMO-554)

This PR adds a management command that can be used to delete messages that have passed an expiration period (defined by `LEARNING_ASSISTANT_MESSAGES_EXPIRY`).


schenedx (approved):
👍

PR #126, opened 2024-11-06
fix: Fixed Couseware History fetching endpoint

rijuma:
When fetching the course history, an error `500` was being produced by the endpoint produced by an `InvalidKeyError`.

This fixes the issue.

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -211,12 +211,12 @@ def save_chat_message(courserun_key, user_id, chat_role, message):
     )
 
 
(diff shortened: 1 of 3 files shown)


alangsto:
Are there any tests that should be updated for these changes?

alangsto (approved):

rijuma:
> Are there any tests that should be updated for these changes?

Indeed, although the tests didn't seem to process the keys in the same way as the live code, since it was not failing. I've updated the tests accordingly. We can discuss it further if you want.

PR #128, opened 2024-11-06
feat: Added timestamp to Learning Assistant History response

rijuma:
In the frontend we are already using the timestamp of the message (although not really showing it).
Instead of making it up on the frontend, I think it's a good moment to add it since it's valuable information for the future.


alangsto (commented):
This looks good to me! If you want to include this in a release, you can add the updated init file and changelog file to this PR. If you don't want to release this right now, then you can just add an entry to the changelog file.

alangsto (approved):

PR #129, opened 2024-11-08
fix: Updated Learning Assistant History payload to return in ascending order

rijuma:
We want the endpoint to return the chat history in ascending order by `timestamp` (creation datetime).

### DIFF
--- a/learning_assistant/__init__.py
+++ b/learning_assistant/__init__.py
@@ -2,6 +2,6 @@
 Plugin for a learning assistant backend, intended for use within edx-platform.
 """
 
-__version__ = '4.4.4'
+__version__ = '4.4.5'
 
 default_app_config = 'learning_assistant.apps.LearningAssistantConfig'  # pylint: disable=invalid-name
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -217,6 +217,10 @@ def get_message_history(courserun_key, user, message_count):
 
     Returns a number of messages equal to the message_count value.
     """
+    # Explanation over the double reverse: This fetches the last message_count elements ordered by creating order DESC.
+    # Slicing the list in the model is an equivalent of adding LIMIT on the query.
+    # The result is the last chat messages for that user and course but in inversed order, so in order to flip them
+    # its first turn into a list and then reversed.
     message_history = list(LearningAssistantMessage.objects.filter(
         course_id=courserun_key, user=user).order_by('-created')[:message_count])[::-1]
     return message_history
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -217,6 +217,6 @@ def get_message_history(courserun_key, user, message_count):
 
     Returns a number of messages equal to the message_count value.
     """
-    message_history = LearningAssistantMessage.objects.filter(
-        course_id=courserun_key, user=user).order_by('-created')[:message_count]
+    message_history = list(LearningAssistantMessage.objects.filter(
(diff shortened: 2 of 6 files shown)


learning_assistant/views.py
```diff
@@ -236,6 +236,6 @@ def get(self, request, course_run_id):
         user = request.user
 
         message_count = int(request.GET.get('message_count', 50))
-        message_history = get_message_history(courserun_key, user, message_count)
+        message_history = reversed(list(get_message_history(courserun_key, user, message_count)))  # Reversing order
```
alangsto, line 239: instead of reversing it here, can we just change the ordering here? https://github.com/edx/learning-assistant/blob/9153aa142435469f0bb14ae75c1f18ab39438f1e/learning_assistant/api.py#L221

rijuma: We can. It was one or the other. I'll make the change.

rijuma: Moved the reverse logic to the API call and updated corresponding tests.

learning_assistant/api.py
```diff
@@ -217,6 +217,6 @@ def get_message_history(courserun_key, user, message_count):
 
     Returns a number of messages equal to the message_count value.
     """
-    message_history = LearningAssistantMessage.objects.filter(
-        course_id=courserun_key, user=user).order_by('-created')[:message_count]
+    message_history = list(LearningAssistantMessage.objects.filter(
+        course_id=courserun_key, user=user).order_by('-created')[:message_count])[::-1]
```
alangsto, line 221: You can get the same affect by doing order_by('created'). Right now you're still sorting twice.

alangsto (changes requested):

learning_assistant/api.py
rijuma: Discussed offline and the alternative would be less performant. I've added a comment with a justification on this.

learning_assistant/api.py
```diff
@@ -217,6 +217,11 @@ def get_message_history(courserun_key, user, message_count):
 
     Returns a number of messages equal to the message_count value.
     """
-    message_history = LearningAssistantMessage.objects.filter(
-        course_id=courserun_key, user=user).order_by('-created')[:message_count]
+    # Explanation over the double reverse: This fetches the last message_count elements ordered by creating order DESC.
+    # The result is the last chat messages for that user and course but in inversed order, so in order to flip them
+    # its first turn into a list and then reversed.
+    # The alternative of getting the last message_count ordering the query in ascending created order is not possible
+    # since negative indexing is not supported.
```
alangsto, line 224: Sorry to nit pick, but I still don't think this explains why we are limiting the number of messages prior to re-ordering them. It would be helpful to explain that the indexing is equivalent to the sql `limit`, which is more efficient than retrieving all messages and then processing that query set.

learning_assistant/__init__.py
```diff
@@ -2,6 +2,6 @@
 Plugin for a learning assistant backend, intended for use within edx-platform.
 """
 
-__version__ = '4.4.4'
+__version__ = '4.4.5'
```
alangsto, line 5: Can you include a changelog entry?

learning_assistant/api.py
rijuma: Just updated it. Let me know what you think.

learning_assistant/__init__.py
rijuma: There is one already: https://github.com/edx/learning-assistant/pull/129/files#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43c

PR #132, opened 2024-11-14
feat: add LearningAssistantAuditTrial model

ilee2u:
https://2u-internal.atlassian.net/browse/COSMO-566

Added a model to track when users start their Xpert LA audit experience (We need to create a new db table to track this because we only store messages up to 30 days, so we can’t simply use message history to find the first message).

LearningAssistantAuditTrial instances will be created on a per user basis, when an audit learner first sends a message using Xpert LA

Learners will have access to the trial across courses.

Each LearningAssistantAuditTrial has:
- User FK (w/ a unique constraint since each user should only have one trial)
- Start date

Part of M1 Backend: https://2u-internal.atlassian.net/wiki/spaces/COSMO/pages/edit-v2/1434320907#M1-Backend


learning_assistant/models.py
```diff
@@ -47,3 +47,18 @@ class LearningAssistantMessage(TimeStampedModel):
...
+
+    A LearningAssistantAuditTrial instance will be created on a per user basis,
+    when an audit learner first sends a message using Xpert LA.
+
+    .. no_pii: This model has no PII.
+    """
+
+    # Unique constraint since each user should only have one trial
+    user = models.ForeignKey(USER_MODEL, db_index=True, on_delete=models.CASCADE, unique=True)
+    start_date = models.DateTimeField(null=True, blank=True)
```
MichaelRoytman, line 64: When would you expect this value to be `null` or blank? If a learner has a record in this database, I'd expect that the `start_date` would be a required field.

The absence of a record in this table would represent a learner without a trial, so I don't see the value in having this be null or blank, but let me know what you think!

MichaelRoytman (changes requested):

learning_assistant/models.py
ilee2u: Ah, that makes sense. I'll make that change real quick.

MichaelRoytman (approved):
Nice!

There’s a mystery `tests/_init__.py` file - not sure if you meant to add that?

PR #134, opened 2024-11-18
feat: add audit gate to CourseChatView

ilee2u:
Ticket: https://2u-internal.atlassian.net/browse/COSMO-568

Add functionality get or create an audit trial if the user is an audit learner.

- If the user’s course enrollment mode (CourseMode) is audit:
  - Get (or create if none exists*) the user’s LearningAssistantAuditTrial (using python API logic created here: https://2u-internal.atlassian.net/browse/COSMO-567 )
  - Get remaining days remaining in their Xpert LA audit trial, using endpoint created above
  - If remaining days is 0, user should not have access and should return 403

- When an audit learner first sends a message in Xpert LA, we’ll need to create a LearningAssistantAuditTrial to save the start date for the audit trial

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -240,20 +240,15 @@ def check_if_audit_trial_is_expired(user, upgrade_deadline):
     if DAYS_SINCE_UPGRADE_DEADLINE >= timedelta(days=0):
         return True
 
-    audit_trial, created = LearningAssistantAuditTrial.objects.get_or_create(
+    audit_trial, _ = LearningAssistantAuditTrial.objects.get_or_create(
         user=user,
         defaults={
             "start_date": datetime.now(),
         },
     )
 
-    # If the trial was just created, then it definitely isn't expired, so return False
-    if created:
-        return False
-
     # If the user's trial is past its expiry date, return "True" for expired. Else, return False
     DAYS_SINCE_TRIAL_START_DATE = datetime.now() - audit_trial.start_date
-    print("DAYS_SINCE_TRIAL_START_DATE:", DAYS_SINCE_TRIAL_START_DATE)
-    if DAYS_SINCE_TRIAL_START_DATE > timedelta(days=AUDIT_TRIAL_MAX_DAYS):
+    if DAYS_SINCE_TRIAL_START_DATE >= timedelta(days=AUDIT_TRIAL_MAX_DAYS):
         return True
     return False
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -2,6 +2,7 @@
 Library for the learning_assistant app.
 """
 import logging
+from datetime import datetime, timedelta
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -10,9 +11,7 @@ from edx_django_utils.cache import get_cache_key
 from jinja2 import BaseLoader, Environment
 from opaque_keys import InvalidKeyError
 
-from datetime import datetime
-
-from learning_assistant.constants import AUDIT_TRIAL_MAX_DAYS, ACCEPTED_CATEGORY_TYPES, CATEGORY_TYPE_MAP
+from learning_assistant.constants import ACCEPTED_CATEGORY_TYPES, AUDIT_TRIAL_MAX_DAYS, CATEGORY_TYPE_MAP
 from learning_assistant.data import LearningAssistantCourseEnabledData
 from learning_assistant.models import (
     LearningAssistantAuditTrial,
@@ -232,18 +231,23 @@ def get_message_history(courserun_key, user, message_count):
     return message_history
 
 
-def check_if_audit_trial_is_expired(user_id):
+def check_if_audit_trial_is_expired(user):
     """
-    Given a user (User), get the corresponding LearningAssistantAuditTrial trial object,
-    or create one if one does not exist yet.
+    Given a user (User), get or create the corresponding LearningAssistantAuditTrial trial object.
     """
-    audit_trial, created = LearningAssistantAuditTrial.objects.get_or_create(user_id)
+    audit_trial, created = LearningAssistantAuditTrial.objects.get_or_create(
+        user=user,
+        defaults={
+            "start_date": datetime.now(),
+        },
+    )
 
     # If the trial was just created, then it definitely isn't expired, so return False
     if created:
         return False
 
     # If the user's trial is expired, return True. Else, return False
-    if (datetime.now() - audit_trial.start_date) < AUDIT_TRIAL_MAX_DAYS:
+    DAYS_SINCE_TRIAL_START_DATE = datetime.now() - audit_trial.start_date
+    if DAYS_SINCE_TRIAL_START_DATE > timedelta(days=AUDIT_TRIAL_MAX_DAYS):
         return True
     return False
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -231,7 +231,7 @@ def get_message_history(courserun_key, user, message_count):
     return message_history
 
 
-def check_if_audit_trial_is_expired(user, upgrade_deadline):
+def audit_trial_is_expired(user, upgrade_deadline):
     """
     Given a user (User), get or create the corresponding LearningAssistantAuditTrial trial object.
     """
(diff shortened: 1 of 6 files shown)


ilee2u:
Reference for modes for audit & verified learners, per the `course_modes` app in edx-platform/common/djangoapps:

AUDIT_MODES = [
    AUDIT,
    HONOR,
    UNPAID_EXECUTIVE_EDUCATION,
    UNPAID_BOOTCAMP
]

VERIFIED_MODES = [
    VERIFIED,
    PROFESSIONAL,
    MASTERS,
    EXECUTIVE_EDUCATION,
    PAID_EXECUTIVE_EDUCATION,
    PAID_BOOTCAMP
]

learning_assistant/api.py
```diff
@@ -224,3 +229,25 @@ def get_message_history(courserun_key, user, message_count):
...
+    """
+    audit_trial, created = LearningAssistantAuditTrial.objects.get_or_create(
+        user=user,
+        defaults={
+            "start_date": datetime.now(),
+        },
+    )
+
+    # If the trial was just created, then it definitely isn't expired, so return False
+    if created:
```
varshamenon4, line 246: Is it possible that a Trial could be created with 0 days left? For example, if the upgrade deadline has already passed?

learning_assistant/views.py
```diff
@@ -68,18 +69,24 @@ def post(self, request, course_run_id):
...
+        # If user does not have a verified enrollment record, or is not staff, they should not have full access
         user_role = get_user_role(request.user, courserun_key)
         enrollment_object = CourseEnrollment.get_enrollment(request.user, courserun_key)
         enrollment_mode = enrollment_object.mode if enrollment_object else None
         if (
-            (enrollment_mode not in CourseMode.VERIFIED_MODES)
+            # NOTE: Will there ever be a case where the user has a course mod that's
+            # in neither VERIFIED_MODES nor AUDIT_MODES that we need to worry about?
+            enrollment_mode not in CourseMode.VERIFIED_MODES
+            and enrollment_mode in CourseMode.AUDIT_MODES
```
varshamenon4, line 80: I'm not positive, but the logic here seems to match: https://github.com/openedx/frontend-app-learning/blob/6813872dd3b6737a1a6e13764feb9fa7596cbe3d/src/courseware/course/chat/Chat.jsx#L43. So I think that this should cover it?

learning_assistant/api.py
```diff
@@ -224,3 +229,25 @@ def get_message_history(courserun_key, user, message_count):
...
+        return False
+
+    # If the user's trial is expired, return True. Else, return False
```
varshamenon4, line 249: Related to above, but we may need to include logic for checking if the course upgrade deadline has passed (unless that's somewhere else, in which case, disregard!)

varshamenon4 (commented):
Looks good so far! Added some comments.

learning_assistant/views.py
ilee2u: If we want to display the learning assistant to audit learners, doesn't that mean we'll need to change that logic in the frontend as well?

MichaelRoytman: The list of [VERIFIED_MODES](https://github.com/openedx/frontend-app-learning/blob/913c8e408662ea5675ed8a27f19afa56b989e44e/src/constants.ts#L40) in the learning MFE is slightly different than the [list in the LMS](https://github.com/openedx/edx-platform/blob/e9ab422101f6e20b07d762b29c75029ffbc2c0b7/common/djangoapps/course_modes/models.py#L208).

I think what is important is that a learner can upgrade from any of the audit modes, right? We'd only want to shower learners than can upgrade the trial. I'm not entirely sure what that exact list is. We may need to figure that out.

ilee2u: It seems like the [VERIFIED_MODES](https://github.com/openedx/frontend-app-learning/blob/913c8e408662ea5675ed8a27f19afa56b989e44e/src/constants.ts#L40) in the learning MFE includes all the modes that are not part of the [AUDIT_MODES in the LMS](https://github.com/openedx/edx-platform/blob/e9ab422101f6e20b07d762b29c75029ffbc2c0b7/common/djangoapps/course_modes/models.py#L208), so I'm wondering we should do one of two things:
- Only check is the user's enrollment mode is in AUDIT_MODES,
- or check that the mode is in frontend-app-learning's version of VERIFIED_MODES.

Kind of leaning towards the first option, but let me know your thoughts.

MichaelRoytman: I tried to find where we allow learners to upgrade and what modes we allow a learner to be in to upgrade. I can't find a singular place. There is the [is_mode_upgradeable](https://github.com/openedx/edx-platform/blob/e9ab422101f6e20b07d762b29c75029ffbc2c0b7/common/djangoapps/course_modes/models.py#L645) method in the LMS, but it's not that widely used.

There is [serialize_upgrade_info](https://github.com/openedx/edx-platform/blob/82297fa41489d4072d49852d512380c8fc251beb/openedx/core/djangoapps/courseware_api/utils.py#L51), which is used by several of the endpoints the learning MFE uses. [can_show_verified_upgrade](https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/courseware/utils.py#L85) uses [is_mode_upsellable](https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/courseware/utils.py#L55), which refers to `CourseMode.UPSELL_TO_VERIFIED_MODES`, as does `can_show_verified_upgrade` later in teh function. [CourseMode.UPSELL_TO_VERIFIED_MODES](https://github.com/openedx/edx-platform/blob/master/common/djangoapps/course_modes/models.py#L230-L231)
is `[HONOR, AUDIT]`.

This kind of makes sense to me. I don't know if we enroll can enroll learners in ExecEd and Bootcamps via the usual upgrade process, so it's possible learners in the `UNPAID_EXECUTIVE_EDUCATION`, and `UNPAID_BOOTCAMP` modes do not see any upgrade messaging.

If that's the case, we may have three states.
1. learners in verified modes and staff - full access
2. learners in the `audit` and `honor` modes - ability to start trial
3. learners in other modes - no access

I'm harping on this because I wouldn't want to advertise a feature that a learner cannot access via upgrade. I think this is worth a conversation at tomorrow's stand up. What do you think?

P.S. I would also want to confirm that we want all "verified" modes to have access to Xpert. Do we want Masters users to have access to Xpert, for example?

learning_assistant/api.py
```diff
@@ -224,3 +229,26 @@ def get_message_history(courserun_key, user, message_count):
     message_history = list(LearningAssistantMessage.objects.filter(
         course_id=courserun_key, user=user).order_by('-created')[:message_count])[::-1]
     return message_history
+
+
+def check_if_audit_trial_is_expired(user, upgrade_deadline):
```
schenedx, line 234: #NIT the function name is verbose. What about `audit_trial_is_expired`?

learning_assistant/api.py
```diff
@@ -224,3 +229,26 @@ def get_message_history(courserun_key, user, message_count):
...
+    audit_trial, _ = LearningAssistantAuditTrial.objects.get_or_create(
+        user=user,
+        defaults={
+            "start_date": datetime.now(),
+        },
+    )
+
+    # If the user's trial is past its expiry date, return "True" for expired. Else, return False
+    DAYS_SINCE_TRIAL_START_DATE = datetime.now() - audit_trial.start_date
+    if DAYS_SINCE_TRIAL_START_DATE >= timedelta(days=AUDIT_TRIAL_MAX_DAYS):
```
schenedx, line 252: What about just do `return DAYS_SINCE_TRIAL_START_DATE >= timedelta(days=AUDIT_TRIAL_MAX_DAYS)`?

learning_assistant/views.py
```diff
@@ -68,18 +69,28 @@ def post(self, request, course_run_id):
...
             and not user_role_is_staff(user_role)
         ):
-            return Response(
-                status=http_status.HTTP_403_FORBIDDEN,
-                data={'detail': 'Must be staff or have valid enrollment.'}
-            )
+            # TODO: Add logic to make sure upgrade deadline has not passed.
```
schenedx, line 83: The TODO comment should not be needed?

schenedx (commented):
See my inline comments

learning_assistant/views.py
ilee2u: I did some digging through the code you linked. Long story short, this seems like what we need (this is a slight modification of what you posted):

1. Any learners in verified modes (per the list in the frontend) or staff will have full access.
2. Any learners with course modes in [CourseMode.UPSELL_TO_VERIFIED_MODES](https://github.com/openedx/edx-platform/blob/master/common/djangoapps/course_modes/models.py#L230-L231) will start or continue their trial (unless it's expired, in which case return 403).
3. Learners in all other course modes (captured in an else statement) will get returned 403.

I've just committed changes reflecting this.

schenedx (approved):
👍

PR #136, opened 2024-11-19
docs: add local setup to readme

varshamenon4:
This PR adds local setup instructions to the readme.


README.rst
```diff
@@ -34,6 +34,61 @@ One Time Setup
   # Here's how you might do that if you have virtualenvwrapper setup.
   mkvirtualenv -p python3.8 learning-assistant
 
+In your ``private.txt`` requirements file in edx-platform, add:
+
+.. code-block::
+
+  -e /edx/src/learning-assistant
+
+In your ``private.py`` settings file in edx-platform, add the below settings. The value of the API key shouldn't matter, because it's not being used at this point, but the setting needs to be there.
```
MichaelRoytman, line 43: Can you specify `lms/envs/private.py`, please? The `private.py` file does not come with the repository, so a user has to add it themselves. If they don't already have this file, it may be unclear where the file should be.

README.rst
```diff
@@ -34,6 +34,61 @@ One Time Setup
   # Here's how you might do that if you have virtualenvwrapper setup.
   mkvirtualenv -p python3.8 learning-assistant
 
+In your ``private.txt`` requirements file in edx-platform, add:
```
MichaelRoytman, line 37: Can you specify `requirements/edx/private.txt` to be clear which file this is, please?

MichaelRoytman, line 37: I think we need to specify somewhere that this repository should be cloned into the `../src` directory relative to the user's `devstack` directory. This could be mentioned in the "One Time Setup" paragraph.

MichaelRoytman, line 37: I think it would be helpful to enumerate which services this plugin depends on - the LMS and discovery. Both you and Marcos ran into issues with the discovery container not running/working, which led to issues reading from the cache. If we instruct users to have both running, that can help avoid issues.

README.rst
```diff
@@ -34,6 +34,27 @@ One Time Setup
...
+
+.. code-block::
+
+  -e /edx/src/learning-assistant
+
+In your ``private.py`` settings file in edx-platform, add the below settings. The value of the API key shouldn't matter, because it's not being used at this point, but the setting needs to be there.
+
+.. code-block::
+
+  CHAT_COMPLETION_API = ''
```
alangsto, line 47: Can you add a comment here that this needs to have a valid URL? And should be copied from edx-internal

README.rst
```diff
@@ -34,6 +34,27 @@ One Time Setup
...
+  CHAT_COMPLETION_API_KEY = ''
+
+  LEARNING_ASSISTANT_PROMPT_TEMPLATE = ''
```
alangsto, line 50: This should also have a comment specifying the value should be copied from edx-internal.

README.rst
```diff
@@ -34,6 +34,27 @@ One Time Setup
...
+.. code-block::
+
+  -e /edx/src/learning-assistant
+
+In your ``private.py`` settings file in edx-platform, add the below settings. The value of the API key shouldn't matter, because it's not being used at this point, but the setting needs to be there.
+
+.. code-block::
+
+  CHAT_COMPLETION_API = ''
+  CHAT_COMPLETION_API_KEY = ''
```
alangsto, line 48: This also needs a value, although it's value doesn't matter at this point.

README.rst
varshamenon4: Yes, good call! Added as a comment above the cloning instructions.

README.rst
varshamenon4: Ah yes! Added as a comment.

README.rst
varshamenon4: Ah makes sense! Added a comment to code block.

README.rst
varshamenon4: Cool! Added comment to code block.

schenedx (approved):
👍

PR #137, opened 2024-11-22
fix: Gates the chat history endpoint behind a waffle flag

rijuma:
Currently, the only logic for the chat history behind the feature flag `learning_assistant.enable_chat_history` is the saving of the messages.

If we want to disable the feature, the chats are no longer being saved but the existing ones will still be returned by the learning Assistant Message History endpoint.

This update makes sure that if the flag for the chat history is disabled, the endpoint would just return an empty list.

### DIFF
--- a/learning_assistant/views.py
+++ b/learning_assistant/views.py
@@ -247,6 +247,10 @@ class LearningAssistantMessageHistoryView(APIView):
                 data={'detail': 'Learning assistant not enabled for course.'}
             )
 
+        # If chat history is disabled, we return no messages as response.
+        if not chat_history_enabled(courserun_key):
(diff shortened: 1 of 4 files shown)


learning_assistant/views.py
```diff
@@ -247,6 +247,10 @@ def get(self, request, course_run_id):
                 data={'detail': 'Learning assistant not enabled for course.'}
             )
 
+        # If chat history is disabled, we return no messages as response.
+        if not chat_history_enabled(courserun_key):
```
varshamenon4, line 251: Nice!

varshamenon4 (approved):
lgtm!

PR #140, opened 2024-12-02
Add a BFFE endpoint for Learning Assistant (LA), including information about LA is enabled, LA message history, and LA audit trial.

MichaelRoytman:
This pull request adds a back-end-for-frontend (BFFE) endpoint for the Learning Assistant to get all the necessary data it needs to function.

The response from this endpoint includes the following information.

* whether the Learning Assistant is enabled
* message history information, if the learner is eligible to use the Learning Assistant
* audit trial information


learning_assistant/views.py
```diff
@@ -271,3 +288,125 @@ def get(self, request, course_run_id):
...
+
+        # Get audit trial.
+        trial = get_audit_trial(user)
+
+        trial_data = {}
+        if trial:
+            trial_data['start_date'] = trial.start_date
+            trial_data['expiration_date'] = trial.expiration_date
+
+        data['trial'] = trial_data
```
ilee2u, line 410: I currently have the frontend extracting 'audit_trial' and not 'trial' from the GET request data. Wondering if we should name this attribute 'audit_trial', or extract the 'trial' attribute in the frontend. Let me know your thoughts!

rijuma: Agree on the question. My vote is on `audit_trial` for consistency.

learning_assistant/urls.py
```diff
@@ -24,4 +29,9 @@
         LearningAssistantMessageHistoryView.as_view(),
         name='message-history',
     ),
+    re_path(
+        fr'learning_assistant/v1/course_id/{COURSE_ID_PATTERN}/summary',
+        LearningAssistantSummaryView.as_view(),
+        name='summary',
+    ),
```
rijuma, line 36: I think we have plans to enable (and integrate) the Unit Summary to Xpert. Maybe this `/summary` reference could be confusing. Maybe we could use `/chat` instead?

varshamenon4 (commented):
Looks great! A few comments and some nits (feel free to disregard those though!).

learning_assistant/urls.py
MichaelRoytman: To clarify, are you suggesting that I change the URL/name to `chat`? I think that would conflict with the chat endpoint, so I would prefer not to do that. I also don't think `chat` represents the purpose of this endpoint.

Personally, I think `summary` and `unit_summary` would be reasonable URLs to coexist, and I wouldn't find them confusing. I don't think we should name the URL for unit summaries `summary`.

What about one of these? `overview`, `details`, or `info`?

learning_assistant/views.py
MichaelRoytman: Sure, no problem. I chose `trial` because the trial is available to more than just learners in the `audit` track, but it's not an issue to change.

learning_assistant/data.py
```diff
@@ -13,3 +15,14 @@ class LearningAssistantCourseEnabledData:
 
     course_key: CourseKey = field(validator=validators.instance_of(CourseKey))
     enabled: bool = field(validator=validators.instance_of(bool))
+
+
+@frozen
```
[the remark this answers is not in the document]
MichaelRoytman: Yes. The `frozen` decorator makes `LearningAssistantAuditTrialData` class immutable, so once an instance of this class is created, its attributes cannot be changed.

learning_assistant/views.py
```diff
@@ -39,6 +43,19 @@
...
+    Path: /learning_assistant/v1/course_id/{course_run_id}
+
+    Parameters:
+        * course_run_id: the ID of the course
+
+    Responses:
+        * 200: OK
+        * 400: Malformed Request - Course ID is not a valid course ID.
+        * 403: Forbidden - Learning assistant not enabled for course or learner does not have a valid enrollment or is
+                           not staff.
```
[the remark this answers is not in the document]
MichaelRoytman: I'm not sure if you're seeing something different than what I'm seeing, but I tend to align subsequent lines of paragraphs to the beginning character of the paragraph, because I find it easier to read. Would you prefer a different indentation?

learning_assistant/api.py
```diff
@@ -231,22 +237,98 @@ def get_message_history(courserun_key, user, message_count):
...
+    """
+    Given a user (User), get or create the corresponding LearningAssistantAuditTrial trial object.
+    """
+    course_mode = CourseMode.objects.get(course=courserun_key)
+
+    upgrade_deadline = course_mode.expiration_datetime()
+
+    # If the upgrade deadline has passed, return True for expired. Upgrade deadline is an optional attribute of a
+    # CourseMode, so if it's None, then do not return True.
+    days_since_upgrade_deadline = datetime.now() - upgrade_deadline if upgrade_deadline else None
```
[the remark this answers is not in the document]
MichaelRoytman: Thanks for the suggestion. I'll make that change.

learning_assistant/api.py
```diff
@@ -224,3 +229,54 @@ def get_message_history(courserun_key, user, message_count):
...
+    """
+    default_trial_length_days = 14
+
+    trial_length_days = getattr(settings, 'LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS', default_trial_length_days)
+
+    if trial_length_days is None:
+        trial_length_days = default_trial_length_days
+
+    # If LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS is set to a negative number, assume it should be 0.
+    # pylint: disable=consider-using-max-builtin
```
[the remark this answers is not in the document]
rijuma: _"I'm ~asking~ telling you to consider, and I won't take a NO for an answer.."_ 🤌

learning_assistant/data.py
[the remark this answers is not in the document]
rijuma: ... just let it go.

learning_assistant/urls.py
rijuma: I totally forgot about the existing `/chat` one. I think `info` fits better, but it's your call.

learning_assistant/views.py
```diff
@@ -271,3 +288,125 @@ def get(self, request, course_run_id):
...
+
+    This endpoint returns all the data necessary for the Learning Assistant to function, including the following
+    information.
+        * whether the Learning Assistant is enabled
+        * message history information, if the learner is eligible to use the Learning Assistant
+        * audit trial information
+
+    Accepts: [GET]
+
+    Path: /learning_assistant/v1/course_id/{course_run_id}/
```
rijuma, line 305: nit: Shouldn't this path include the `/summary` part (or whatever variant you choose for it)?

MichaelRoytman: Yes! Thank you for noticing that - very observant 😀.

ilee2u (approved):

PR #143, opened 2024-12-10
Add  audit_trial_length_days attribute to ChatSummaryView response.

MichaelRoytman:
### Description

**Jira**: [COSMO-603](https://2u-internal.atlassian.net/browse/COSMO-603)

In order for the frontend to know the trial length of an audit trial, a `audit_trial_length` attribute is added to the `ChatSummaryView` response as an attribute of the `audit_trial` object.

This property is supplied whether or not a learner has an audit trial. It represents the number of days in an audit trial as currently configured. It does not represent the number of days in the learner's current trial.

### DIFF
--- a/learning_assistant/utils.py
+++ b/learning_assistant/utils.py
@@ -112,3 +112,28 @@ def user_role_is_staff(role):
     * bool: whether the user's role is that of a staff member
     """
     return role in ('staff', 'instructor')
+
+
+def get_audit_trial_length_days():
+    """
+    Return the length of an audit trial in days.
+
+    Arguments:
+    * None
+
+    Returns:
+    * int: the length of an audit trial in days
+    """
+    default_trial_length_days = 14
+
+    trial_length_days = getattr(settings, 'LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS', default_trial_length_days)
+
+    if trial_length_days is None:
+        trial_length_days = default_trial_length_days
+
+    # If LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS is set to a negative number, assume it should be 0.
+    # pylint: disable=consider-using-max-builtin
+    if trial_length_days < 0:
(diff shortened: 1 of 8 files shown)


learning_assistant/views.py
```diff
@@ -402,7 +402,9 @@ def get(self, request, course_run_id):
         # Get audit trial.
         trial = get_audit_trial(user)
 
-        trial_data = {}
+        trial_data = {
```
MichaelRoytman, line 405: Note that this means that `audit_trial` will never be an empty dictionary, since it'll always have the value of `trial_length_days`.

learning_assistant/utils.py
```diff
@@ -112,3 +112,28 @@ def user_role_is_staff(role):
     * bool: whether the user's role is that of a staff member
     """
     return role in ('staff', 'instructor')
+
+
+def get_audit_trial_length_days():
```
varshamenon4, line 117: Nice idea to move this!

learning_assistant/views.py
varshamenon4: Is there any implication for this never being empty?

tests/test_api.py
```diff
@@ -492,22 +492,24 @@ class GetAuditTrialExpirationDateTests(TestCase):
     Test suite for get_audit_trial_expiration_date.
```
varshamenon4, line 492: Nice updates to tests!

varshamenon4 (approved):
Had a question but not blocking... lgtm! Approved!

learning_assistant/views.py
MichaelRoytman: I don't think so, but let me check with the team.

MichaelRoytman: I confirmed there's no issue with the team, so I'll go ahead and merge this.

varshamenon4: Thanks for checking!

PR #146, opened 2024-12-18
fix: use course enrollment instead of course mode to calculate audit trial is expired

varshamenon4:
[COSMO-616](https://2u-internal.atlassian.net/browse/COSMO-616)

As I was trying to understand how to set the upgrade deadline (i.e. the topic in standup about prioritizing audit vs. honor), I ran into some questions around CourseEnrollment and CourseMode which might change how [audit_trial_is_expired](https://github.com/edx/learning-assistant/blob/39d2f2185e60b2fe17585f1b636f060c75d20eef/learning_assistant/api.py#L310) is implemented.

**Context**

Currently [audit_trial_is_expired](https://github.com/edx/learning-assistant/blob/39d2f2185e60b2fe17585f1b636f060c75d20eef/learning_assistant/api.py#L310) gets the CourseMode (this is where it gets multiple) using the courserun_key and then uses the expiration_datetime (aka upgrade_deadline) to calculate whether the audit trial is expired

But, the CourseMode [expiration_datetime](https://github.com/openedx/edx-platform/blob/882475dd5e81df717100ae3e54f838c326370494/common/djangoapps/course_modes/models.py#L74) corresponds to the expiration of this course mode. So if "For example, if there is a verified mode that expires on 1/1/2015, then users will be able to upgrade into the verified mode before that date. Once the date passes, users will no longer be able to enroll as verified." This means that for the purpose of the [audit_trial_is_expired](https://github.com/edx/learning-assistant/blob/39d2f2185e60b2fe17585f1b636f060c75d20eef/learning_assistant/api.py#L310), we want to get the upgrade deadline for the verified mode, not for the audit or honor mode which is what we would be doing now, since the learner is an audit learner

**Proposed Solution**

I propose that we use the CourseEnrollment [upgrade_deadline](https://github.com/openedx/edx-platform/blob/master/common/djangoapps/student/models/course_enrollment.py#L1248) to get the upgrade deadline for the course.

This method gets the verified modes (if there are any) and calculates the upgrade deadline, which is what we need for this method.

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -3,7 +3,7 @@ Library for the learning_assistant app.
 """
 import datetime
 import logging
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -11,11 +11,6 @@ from django.core.cache import cache
 from edx_django_utils.cache import get_cache_key
 from jinja2 import BaseLoader, Environment
 from opaque_keys import InvalidKeyError
-
-try:
-    from common.djangoapps.course_modes.models import CourseMode
-except ImportError:
-    CourseMode = None
 
 from learning_assistant.constants import ACCEPTED_CATEGORY_TYPES, CATEGORY_TYPE_MAP
 from learning_assistant.data import LearningAssistantAuditTrialData, LearningAssistantCourseEnabledData
@@ -307,19 +302,25 @@ def get_or_create_audit_trial(user):
     )
 
 
-def audit_trial_is_expired(audit_trial_data, courserun_key):
+def audit_trial_is_expired(enrollment, audit_trial_data):
     """
-    Given a user (User), get or create the corresponding LearningAssistantAuditTrial trial object.
-    """
-    course_mode = CourseMode.objects.get(course=courserun_key)
+    Given an enrollment and audit_trial_data, return whether the audit trial is expired as a boolean.
 
-    upgrade_deadline = course_mode.expiration_datetime()
+    Arguments:
+    * enrollment (CourseEnrollment): the user course enrollment
+    * audit_trial_data (LearningAssistantAuditTrialData): the data related to the audit trial
+
+    Returns:
+    * audit_trial_is_expired (boolean): whether the audit trial is expired
+    """
+    upgrade_deadline = enrollment.upgrade_deadline
+    today = datetime.now(tz=timezone.utc)
 
     # If the upgrade deadline has passed, return True for expired. Upgrade deadline is an optional attribute of a
-    # CourseMode, so if it's None, then do not return True.
-    days_until_upgrade_deadline = datetime.now() - upgrade_deadline if upgrade_deadline else None
+    # CourseEnrollment, so if it's None, then do not return True.
+    days_until_upgrade_deadline = today - upgrade_deadline if upgrade_deadline else None
     if days_until_upgrade_deadline is not None and days_until_upgrade_deadline >= timedelta(days=0):
         return True
 
     # If the user's trial is past its expiry date, return True for expired. Else, return False.
-    return audit_trial_data is None or audit_trial_data.expiration_date <= datetime.now()
+    return audit_trial_data is None or audit_trial_data.expiration_date <= today
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -2,7 +2,7 @@
 Test cases for the learning-assistant api module.
 """
 import itertools
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from unittest.mock import MagicMock, patch
 
 import ddt
@@ -491,6 +491,7 @@ class GetAuditTrialExpirationDateTests(TestCase):
     """
     Test suite for get_audit_trial_expiration_date.
     """
+
     @ddt.data(
         (datetime(2024, 1, 1, 0, 0, 0), datetime(2024, 1, 2, 0, 0, 0), 1),
         (datetime(2024, 1, 18, 0, 0, 0), datetime(2024, 1, 19, 0, 0, 0), 1),
@@ -516,6 +517,7 @@ class GetAuditTrialTests(TestCase):
     """
     Test suite for get_audit_trial.
     """
+
     @freeze_time('2024-01-01')
     def setUp(self):
         super().setUp()
@@ -548,6 +550,7 @@ class GetOrCreateAuditTrialTests(TestCase):
     """
     Test suite for get_or_create_audit_trial.
     """
+
     def setUp(self):
         super().setUp()
         self.user = User(username='tester', email='tester@test.com')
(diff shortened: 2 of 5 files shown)


learning_assistant/api.py
```diff
@@ -311,9 +316,9 @@ def audit_trial_is_expired(audit_trial_data, courserun_key):
     """
     Given a user (User), get or create the corresponding LearningAssistantAuditTrial trial object.
     """
-    course_mode = CourseMode.objects.get(course=courserun_key)
+    course_enrollment = CourseEnrollment.objects.get(user=audit_trial_data.user, course=courserun_key)
 
-    upgrade_deadline = course_mode.expiration_datetime()
+    upgrade_deadline = course_enrollment.upgrade_deadline()
```
alangsto, line 321: Nice! This seems like the right approach

learning_assistant/api.py
```diff
@@ -16,6 +16,11 @@
     from common.djangoapps.course_modes.models import CourseMode
 except ImportError:
     CourseMode = None
+# todo: do we need try / catch for this?
+try:
+    from common.djangoapps.course_modes.models import CourseEnrollment
```
alangsto, line 21: I see that this wasn't done for the CourseMode import, but we have a file here https://github.com/edx/learning-assistant/blob/main/learning_assistant/platform_imports.py that can be used for platform imports. Using this pattern makes mocking these classes/functions a bit easier in unit testing.

tests/test_api.py
```diff
@@ -599,12 +599,11 @@ def setUp(self):
...
-    @patch('learning_assistant.api.CourseMode')
-    def test_upgrade_deadline_expired(self, mock_course_mode):
-
-        mock_mode = MagicMock()
-        mock_mode.expiration_datetime.return_value = datetime.now() - timedelta(days=1)  # yesterday
-        mock_course_mode.objects.get.return_value = mock_mode
+    @patch('learning_assistant.api.CourseEnrollment')
+    def test_upgrade_deadline_expired(self, mock_course_enrollment):
+        # mock_enrollment = MagicMock(upgrade_deadline=)
+        mock_course_enrollment.return_value.upgrade_deadline.return_value = datetime.now() - timedelta(days=1)  # yesterday
```
varshamenon4, line 605: this returns a MagicMock object instead of a datetime... not sure why! Spent a bit of time and referenced this: https://stackoverflow.com/questions/38199008/python-returns-magicmock-object-instead-of-return-value

the signature looks like this: `def audit_trial_is_expired(enrollment, audit_trial_data)`

And the call to get the upgrade_deadline is this: upgrade_deadline = enrollment.upgrade_deadline. 

I wonder if it is weird because I'm mocking the course enrollment when it is a param rather than the other similar cases where the mocked object is retrieved via a getter.

varshamenon4: All resolved! just needed to remove the return value.

learning_assistant/api.py
```diff
@@ -307,19 +302,25 @@ def get_or_create_audit_trial(user):
...
+    * audit_trial_data (LearningAssistantAuditTrialData): the data related to the audit trial
 
-    upgrade_deadline = course_mode.expiration_datetime()
+    Returns:
+    * audit_trial_is_expired (boolean): whether the audit trial is expired
+    """
+    upgrade_deadline = enrollment.upgrade_deadline
+    today = date.today()
 
     # If the upgrade deadline has passed, return True for expired. Upgrade deadline is an optional attribute of a
```
varshamenon4, line 319: Note: I've changed the way that this is calculated to avoid issues with timezone naive/aware timezones. This now converts all of the datetimes into dates so that we're just calculating the difference in whole days (I believe this is more consistent with the calculations of days remaining in the frontend as well).

varshamenon4 (commented):
Just added a comment explaining why and what has changed in the audit trial calculation.

MichaelRoytman (commented):
This looks good, but I have some concerns about the time calculations. I tried to be very thorough about my thoughts because I personally find time confusing. Let me know what you think!

The CourseEnrollment upgrade date was a really nice find! 🎉

learning_assistant/api.py
rijuma: I think forcing the timezone as `tz=None` might cause issues, since the dates should be treated in the same timezone.

In the case of the frontend, the date received is in UTC but when a date is created like in the hook as `new Date(auditTrial.expirationDate)`, the resulting date is converted to the local timezone, that's why it can be compared with `Date.now()` since in the frontend, both dates are in the same timezone.

I'm no expert on Python, but I think it should be similar, so parsing a UTC time should (IMHO) be created with the server timezone, so it should be able to be compared with `datetime.now()`

varshamenon4: Good call! I've updated to have the timezone set when creating datetime.now() to have it be timezone aware so it can be compared with the other times.

MichaelRoytman: Do we know anything about the `upgrade_deadline` in terms of the timezone? Is it always in UTC? Or is it in the learner's local server time? Is it an aware `datetime`? I assume so given the need to keep `today` aware.

I'm concerned about the use of `timezone.utc` here, because we may be comparing against another timezone in `upgrade_deadline`, and Python isn't going to automatically convert `today` to the learner's local server time or to the same timezone that `upgrade_deadline` is in.

varshamenon4: I can do some more research on this but the reason I did this was because my understanding is that when comparing two aware datetimes, the datetime is considered. So you can compare two aware datetimes and the timezone is taken into account. See here: https://stackabuse.com/comparing-datetimes-in-python-with-and-without-timezones/.

If it's better I could convert all the timezones to UTC but I didn't think this was necessary.

schenedx (approved):
👍

learning_assistant/api.py
rijuma: Didn't know that. Good call.

rijuma (approved):

learning_assistant/api.py
MichaelRoytman: I think you're right that Python compares them correctly under-the-hood. I just couldn't find that state explicitly in the documentation. Looks good to me!

MichaelRoytman (approved):

learning_assistant/api.py
varshamenon4: Yeah, actually I'm not sure as I'm testing so may just need to convert first.

varshamenon4: So yes, I confirmed by testing this:
```
@freeze_time('2024-01-01 00:00:01')
    def test_timezone(self):
        today_utc = datetime.now(tz=timezone.utc)
        london = datetime.now(zoneinfo.ZoneInfo("Europe/London"))
        nyc = datetime.now(zoneinfo.ZoneInfo("America/New_York"))
        logging.info(today_utc)
        logging.info(london)
        logging.info(nyc)

        self.assertEqual(today_utc, london)
        self.assertNotEqual(today_utc, nyc)
        self.assertNotEqual(london, nyc)
```

And these are the outputs of each:
```
FakeDatetime(2024, 1, 1, 0, 0, 1, tzinfo=datetime.timezone.utc)
FakeDatetime(2024, 1, 1, 0, 0, 1, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
FakeDatetime(2023, 12, 31, 19, 0, 1, tzinfo=zoneinfo.ZoneInfo(key='America/New_York'))
```

So it looks like the comparison of datetimes takes the timezones into account!

PR #147, opened 2024-12-18
fix: correct params for expiration_date

ilee2u:
Correction on the parameters for the expiration_date in https://github.com/edx/learning-assistant/pull/145, which is causing the table to error.


schenedx (approved):
👍

PR #149, opened 2024-12-18
docs: update setup instructions

alangsto:
Updated setup instructions to include directions for setting up Xpert for audit learners locally.


schenedx (approved):
👍

PR #152, opened 2025-01-06
feat: gate use of v2 endpoint with waffle flag

alangsto:
## [COSMO-620](https://2u-internal.atlassian.net/browse/COSMO-620)

The Xpert platform team has released a new v2 endpoint for processing messages. This PR gates use of the v2 endpoint by a new waffle flag. 

I also cleaned up some code related to an unused API key and an unused waffle flag related to course content.


learning_assistant/toggles.py
```diff
@@ -37,15 +36,15 @@ def _is_learning_assistant_waffle_flag_enabled(flag_name, course_key):
...
     """
-    Return whether the learning_assistant.enable_course_content WaffleFlag is on.
+    Return whether the learning_assistant.enable_chat_history WaffleFlag is on.
     """
-    return _is_learning_assistant_waffle_flag_enabled(ENABLE_COURSE_CONTENT, course_key)
+    return _is_learning_assistant_waffle_flag_enabled(ENABLE_CHAT_HISTORY, course_key)
 
 
-def chat_history_enabled(course_key):
+def v2_endpoint_enabled():
```
alangsto, line 46: I did not make this a course waffle flag as I don't believe we need a course by course roll out.

learning_assistant/utils.py
```diff
@@ -47,30 +49,37 @@ def get_reduced_message_list(prompt_template, message_list):
...
 
 
 def get_chat_response(prompt_template, message_list):
     """
     Pass message list to chat endpoint, as defined by the CHAT_COMPLETION_API setting.
     """
-    completion_endpoint = getattr(settings, 'CHAT_COMPLETION_API', None)
-    completion_endpoint_key = getattr(settings, 'CHAT_COMPLETION_API_KEY', None)
-    if completion_endpoint and completion_endpoint_key:
-        headers = {'Content-Type': 'application/json', 'x-api-key': completion_endpoint_key}
```
varshamenon4, line 73: Question: is there no longer the need for an api key with the new endpoint?

learning_assistant/utils.py
```diff
@@ -47,30 +49,37 @@ def get_reduced_message_list(prompt_template, message_list):
...
+    messages = get_reduced_message_list(prompt_template, message_list)
+
     response_body = {
-        'message_list': get_reduced_message_list(prompt_template, message_list),
+        'message_list': [{'role': 'system', 'content': prompt_template}] + messages,
     }
 
+    if v2_endpoint_enabled():
+        response_body = {
+            'client_id': getattr(settings, 'CHAT_COMPLETION_CLIENT_ID', 'edx_olc_la'),
```
varshamenon4, line 67: Question: what is the 'edx_olc_la'? Is there a reason why this is hardcoded rather than added as a setting?

tests/test_utils.py
```diff
@@ -126,18 +145,18 @@ def test_message_list_reduced(self):
         """
         # pass in copy of list, as it is modified as part of the reduction
         reduced_message_list = get_reduced_message_list(self.prompt_template, self.message_list)
-        self.assertEqual(len(reduced_message_list), 2)
+        self.assertEqual(len(reduced_message_list), 1)
```
varshamenon4, line 148: Question: might be missing this in the implementation but can you clarify why the length goes down by 1 for this assert and the assert below?

CHANGELOG.rst
```diff
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 **********
 
+4.6.4 - 2025-01-06
+******************
+* Gate use of the Xpert platform v2 endpoint with a waffle flag.
```
varshamenon4, line 19: Nit: would using the v2 endpoint justify considering this a minor release than a patch? So 4.7.0 instead of 4.6.4? Sort of a nit but I wasn't sure how big of a change this would represent api-wise.

varshamenon4 (commented):
Looks great so far! Just a couple of clarifying questions and nits.

PR #154, opened 2025-01-08
fix: use django utils timezone to make timezone aware

varshamenon4:
This PR fixes a bug which caused an error in audit_trial_is_expired which was happening due to an incompatible comparison in timezone naive and aware datetimes.

Instead of using datetime.now(), we now use timezone.now() from Django utils which makes timezone aware / naive based on the USE_TZ setting. This changes both the instance of this in audit_trial_is_expired and also in the LearningAssistantAuditTrialData's start_date and expiration_date. 

This ensures consistency in the comparisons.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -2,7 +2,7 @@
 Test cases for the learning-assistant api module.
 """
 import itertools
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 from unittest.mock import MagicMock, patch
 
 import ddt
@@ -593,15 +593,15 @@ class CheckIfAuditTrialIsExpiredTests(TestCase):
     Test suite for audit_trial_is_expired.
     """
 
+    @freeze_time('2024-01-01 00:00:01')
     def setUp(self):
         super().setUp()
         self.course_key = CourseKey.from_string('course-v1:edx+fake+1')
         self.user = User(username='tester', email='tester@test.com')
         self.user.save()
 
-    @freeze_time('2024-01-01 00:00:01 UTC')
     def test_upgrade_deadline_expired(self):
-        today = datetime.now(tz=timezone.utc)
+        today = datetime.now()
         mock_enrollment = MagicMock()
         mock_enrollment.upgrade_deadline = today - timedelta(days=1)  # yesterday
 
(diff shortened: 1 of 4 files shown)


tests/test_api.py
```diff
@@ -593,15 +594,15 @@ class CheckIfAuditTrialIsExpiredTests(TestCase):
     Test suite for audit_trial_is_expired.
     """
 
+    @freeze_time('2024-01-01 00:00:01')
```
varshamenon4, line 597: One change of note in the tests. I went back to not including a timezone, so all of the datetimes in the tests are timezone naive. This was how it was prior to the initial bug fix but I had changed it due to the prior implementation.

I changed this because now that I'm using `timezone.now()` instead of `datetime.now()`, there is some weirdness in how the tests work out. Freezegun currently mocks out datetime, but not django utils timezone. This meant that with the timezones intact, the `timezone.now()` [here in audit_trial_is_expired](https://github.com/edx/learning-assistant/pull/154/files#diff-b99ef3507f4b4f0017e8d985196d85876eaf15d920fc11729f8b530e49aaa392R317) was returning a timezone naive datetime, because `USE_TZ` is false. But, the other datetimes set in the test suite were timezone aware. I also tried updating test_settings.py to include `USE_TZ = True`. But this understandably broke many of the tests, due to missing timezones. This could be something we go back and fix, but I think what we need to test here is that `timezone.now()` picks up the timezone / is naive based on the `USE_TZ` setting, so I don't think it is necessary to make everything timezone aware. Rather what is captured in this testing is that there are no inconsistent naive vs aware datetime comparisons.

I removed the timezone, because this is now working as expected... the datetimes are timezone naive and so the `timezone.now()` calculation is also naive. 

Open to feedback but just wanted to note my reasoning behind this change.

MichaelRoytman: That makes sense to me. I think we should run our tests with `USE_TZ=True`, since that's what's set on `edx.org`, but I agree that refactoring the rest of this library's tests to make that wourk could be follow up work for another time and isn't that consequential for this test.

MichaelRoytman (approved):
Looks good to me! I appreciate your persistence in fixing the timezone issue 💪. Just a quick reminder to also update the `CHANGELOG` and `__init__.py`!

tests/test_api.py
varshamenon4: Good call! I will create a follow up ticket so we have it tracked.

varshamenon4: https://2u-internal.atlassian.net/browse/COSMO-628

PR #159, opened 2025-02-05
feat: Updated Chat Trial Audit gating consistency

rijuma:
## Description
Ticket: [COSMO-612 🔒](https://2u-internal.atlassian.net/browse/COSMO-612)

There was some differences between the Xpert Chat Trial conditions for the trial [on the frontend](https://github.com/openedx/frontend-app-learning/pull/1581) vs the backend.

Removed deprecated `enabled` and `history` endpoints that were integrated in the new summary endpoint. (Ticket: [COSMO-600 🔒](https://2u-internal.atlassian.net/browse/COSMO-600))

These changes would make the checks more consistent.


learning_assistant/views.py
```diff
@@ -380,8 +296,19 @@ def get(self, request, course_run_id):
...
+
+        valid_dates = (
+            (start <= today if start else True)
+            and (end >= today if end else True)
+        )
+
         has_trial_access = (
-            enrollment_mode in valid_trial_access_modes
+            valid_dates
+            and enrollment_mode in valid_trial_access_modes
```
MichaelRoytman, line 311: I can't comment on line 318, but could you remove the parentheses in `(enrollment_mode in valid_full_access_modes)`? They're not necessary; I'm not sure why I added them originally.

tests/test_views.py
```diff
@@ -77,6 +97,27 @@ def setUp(self):
         self.user.save()
         self.client.login_user(self.user)
 
+        self.mocked_datetime = patch(
```
MichaelRoytman, line 100: I think there's two better ways to do this.

1. The mock library conveniently has an example of how to make a function of the `datetime` module without a dummy class [here](https://docs.python.org/3/library/unittest.mock-examples.html#partial-mocking)!
2. You could also consider using the `freezegun` library, which we use in several places. [Here's](https://github.com/edx/learning-assistant/blob/93d620e49447c54d6e70f7816c6823dcc5eb2534/tests/test_api.py#L527-L533) an example in this repository.

tests/test_views.py
```diff
@@ -131,6 +166,86 @@ def test_invalid_messages(self, mock_mode, mock_enrollment, mock_get_user_role,
...
+        ]
+
+        response = self.client.post(
+            reverse('chat', kwargs={'course_run_id': self.course_id})+f'?unit_id={test_unit_id}',
+            data=json.dumps(test_data),
+            content_type='application/json'
+        )
+
+        if dates['expects_fail']:
+            self.assertEqual(response.status_code, 403)
```
MichaelRoytman, line 245: The `IsAuthenticated` permission on the view may return a `403`, so this condition may be satisfied by that as well. I'd recommend also checking the `detail` for `Learning assistant not enabled for course.` I've had situations where I thought I was correctly testing this sort of thing, but I was actually hitting an unrelated error earlier in the code because I hadn't set up the test properly.

P.S. I know there are several other places we just check for a `403` in this file, so it makes sense that you did as well.

What do you think?

MichaelRoytman (changes requested):
Looks good! I left a little feedback.

Two more things.
1. I don't think that `chore` is the appropriate commit type here. This is a `feat`, in my opinion, because it changes the functionality of the application. If you agree, could you update the commit message and PR title?
2. Could you add `COSMO-600` to your pull request description and drop a link to this PR on the ticket in Jira, please? It will help us track our work better.

Thanks!

rijuma:
> Looks good! I left a little feedback.
> 
> Two more things.
> 
> 1. I don't think that `chore` is the appropriate commit type here. This is a `feat`, in my opinion, because it changes the functionality of the application. If you agree, could you update the commit message and PR title?
> 2. Could you add `COSMO-600` to your pull request description and drop a link to this PR on the ticket in Jira, please? It will help us track our work better.
> 
> Thanks!

Thank you for your exhaustive review @MichaelRoytman. Let me know what you think on the updates.

tests/test_views.py
rijuma: I agree. Just added the assert to match the returning payload.

learning_assistant/views.py
```diff
@@ -380,8 +296,19 @@ def get(self, request, course_run_id):
...
 
+        course_data = get_cache_course_run_data(course_run_id, ['start', 'end'])
+        today = datetime.now()
+        start = course_data.get('start', None)
+        end = course_data.get('end', None)
+
+        valid_dates = (
+            (start <= today if start else True)
+            and (end >= today if end else True)
+        )
```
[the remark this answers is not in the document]
rijuma: Whoops! Thanks!

learning_assistant/views.py
```diff
@@ -131,7 +133,20 @@ def post(self, request, course_run_id):
...
 
-        if not learning_assistant_enabled(courserun_key):
+        course_data = get_cache_course_run_data(course_run_id, ['start', 'end'])
+        today = datetime.now()
+        start = course_data.get('start', None)
+        end = course_data.get('end', None)
+
+        valid_dates = (
+            (start <= today if start else True)
+            and (end > today if end else True)
```
MichaelRoytman, line 143: I noticed that you removed the `end >= today`. Do you plan to make the same change on the frontend?

(By the way, I do think this is very edge-casey, and on further reflection, isn't a big deal. So feel free to change it back if it makes your life easier.)

MichaelRoytman (approved):
Looks great! Thank you for incorporating my feedback.

I left one comment on the change I originally suggested to the `>= end` condition, but it's not blocking.

learning_assistant/views.py
rijuma: Just rolled it back to match the frontend. To be honest, since both dates includes time it's just pointless to consider the exact moment that both dates are matching. I've rolled it back to be consistent with the frontend.
Giving the extreme edge case, I don't think it's worth the surgical analysis we are putting into.

PR #160, opened 2025-02-05
feat: add optimizely to decide trial length

alangsto:
## [COSMO-643](https://2u-internal.atlassian.net/browse/COSMO-643)

If an optimizely SDK key is available, we should use optimizely to determine how long a user's audit trial should be. In the case that the experiment is not enabled or not available (which would result in the variation key being null), we default to the existing trial length setting.


learning_assistant/views.py
```diff
@@ -410,6 +410,6 @@ def get(self, request, course_run_id):
 
         data['audit_trial'] = trial_data
 
-        data['audit_trial_length_days'] = get_audit_trial_length_days()
+        data['audit_trial_length_days'] = get_audit_trial_length_days(user.id)
```
alangsto, line 413: My one concern with this approach here is that this endpoint is used for both audit and non-audit learners. This means that optimizely will include verified learners in the experiment.

Is the `audit_trial_length_days` field needed on the frontend for non-audit learners?

MichaelRoytman: I believe it's only used if the learner is eligible for the experiment [here](https://github.com/edx/frontend-lib-learning-assistant/blob/c2304e5a6b53429985448fc1634c3c621f8b7391/src/components/Disclosure/index.jsx#L43). But it's a good point that, both on the backend and frontend, a learner in the `verified` track will be included in the experiment.

We should be able to use an Optimizely [audience](https://docs.developers.optimizely.com/feature-experimentation/docs/target-audiences) for this. This would involve setting up a new audience in the Optimizely dashboard and then passing along a value for the attribute (i.e. the learner's course enrollment mode) in the use of the `OptimizelyProvider` on the frontend ([example](https://github.com/edx/frontend-lib-learning-assistant/blob/c2304e5a6b53429985448fc1634c3c621f8b7391/src/experiments/ExperimentsProvider.jsx#L17)) or by [setting user context](https://docs.developers.optimizely.com/feature-experimentation/docs/create-user-context-python#audience-attributes) on the backend. That will exclude learners in the `verified` track from the experiment.

However... will this be an issue of a learner starts in `audit` mode and then upgrades to `verified`? I have to run to a meeting, so I can't check the code for whether this is an issue, but I'll follow up!

MichaelRoytman (commented):
This looks good! I left a few comments. I'll plan to re-review once we iron out what to do about `verified` learners!

learning_assistant/utils.py
```diff
@@ -123,19 +124,43 @@ def user_role_is_staff(role):
...
     """
-    default_trial_length_days = 14
+    variation = get_optimizely_variation(user_id)
 
-    trial_length_days = getattr(settings, 'LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS', default_trial_length_days)
+    if (
+        variation['enabled']
+        and variation['variation_key'] == getattr(settings, 'OPTIMIZELY_LEARNING_ASSISTANT_TRIAL_VARIATION_KEY', '')
+    ):
+        trial_length_days = 28
```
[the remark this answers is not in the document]
alangsto: Just talked about this with you, but I can rename this variable to make it clearer that this variation key should be the `xpert_audit_28_day_trial`!

test_settings.py
```diff
@@ -90,3 +90,7 @@ def root(*args):
 LEARNING_ASSISTANT_AVAILABLE = True
 
 LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS = 14
+
+OPTIMIZELY_FULL_STACK_SDK_KEY = ''
```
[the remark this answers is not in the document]
alangsto: No, but I'm going to include it as part of my work! The SDK key value already exists for edx-platform

learning_assistant/utils.py
```diff
@@ -123,19 +124,43 @@ def user_role_is_staff(role):
...
-def get_audit_trial_length_days():
+def get_optimizely_variation(user_id):
+    """
+    Return whether or not a user is part of an optimizely variation.
+    """
+    if not getattr(settings, 'OPTIMIZELY_FULL_STACK_SDK_KEY', None):
+        enabled = False
+        variation_key = None
+    else:
+        optimizely_client = optimizely.Optimizely(sdk_key=getattr(settings, 'OPTIMIZELY_FULL_STACK_SDK_KEY', ''))
```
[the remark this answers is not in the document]
alangsto: Good catch, we can leave the default out here

learning_assistant/utils.py
```diff
@@ -123,19 +124,43 @@ def user_role_is_staff(role):
...
+        user = optimizely_client.create_user_context(str(user_id))
+        decision = user.decide(getattr(settings, 'OPTIMIZELY_LEARNING_ASSISTANT_TRIAL_EXPERIMENT_KEY', ''))
```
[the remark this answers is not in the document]
alangsto: If `decide` is passed an empty string, a decision object will still be returned, but the `variation_key` will have a null value and `enabled` will be `False`.

learning_assistant/utils.py
```diff
@@ -123,19 +124,43 @@ def user_role_is_staff(role):
...
     Returns:
     * int: the length of an audit trial in days
     """
-    default_trial_length_days = 14
+    variation = get_optimizely_variation(user_id)
 
-    trial_length_days = getattr(settings, 'LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS', default_trial_length_days)
+    if (
+        variation['enabled']
+        and variation['variation_key'] == getattr(settings, 'OPTIMIZELY_LEARNING_ASSISTANT_TRIAL_VARIATION_KEY', '')
```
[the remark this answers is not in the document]
alangsto: If `off`, `control`, and `xpert_audit_14_day_trial` all result in the trial length being 14 days, do they each need a separate case? I can rename the variation key setting, but my thought was that it would be set to `xpert_audit_28_day_trial`, as that's the only key that results in a different behavior.

learning_assistant/views.py
MichaelRoytman: I completely understand what you're saying. However, I am concerned about the quality of the data. Any time we use the Optimizely SDK to make a decision about the learner's status in the experiment, Optimizely counts that as an "impression". Here is some documentation about [impressions](https://docs.developers.optimizely.com/feature-experimentation/docs/impressions-and-decisions#how-the-sdk-triggers-impressions).

I want to point out this paragraph from that same page.

> Impressions generate the visitor counts on the [Experimentation Results](https://docs.developers.optimizely.com/feature-experimentation/docs/analyze-results) page. They also serve as the denominator for measuring conversion rates.

My concern is that, although we can argue that the impact to each variation of deciding to include verified learners in the experiment will be equal, that does not mean that this decision will not affect the data. For example, we may reach statistical significance far earlier than we otherwise would because we're including all of these extra impressions that are meaningless (i.e. for learners that see no experimental code). And there is also no guarantee that we'll get an equal distribution of non-upgrade eligible learners into each of the two variations.

I also don't think we can say that that the impact to each variation will be equal and will be canceled out between the two variations because we're including non-upgrade eligible learners into both variations.

I know we're not running a clinical trial here or anything 😆, but if we can keep the data cleaner and reduce the number of variables, then I think our data will be more trustworthy and easier to interpret. Counting learners who are not actually in the experiment adds additional noise. I would advocate for excluding learners that are not eligible for the trial. I'm pushing for this because I think that it's important that we collect good data as much as we reasonable can, and I don't think it's a difficult thing to add.

Let me know what you think!

test_settings.py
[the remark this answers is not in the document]
MichaelRoytman: Thank you! That's great.

Could you share where it's defined for the platform? I couldn't find it when I ran the search below originally. I only see the frontend setting. Am I missing it? I'd like to know where to look for it in the future.

https://github.com/search?q=org%3Aedx+OPTIMIZELY_FULL_STACK_SDK_KEY&type=code&p=1

learning_assistant/utils.py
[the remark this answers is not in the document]
MichaelRoytman: That sounds good! Thank you!

test_settings.py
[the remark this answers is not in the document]
alangsto: I just realized I had a small typo! But here is a link to where this value is defined in edx-internal: https://github.com/edx/edx-internal/blob/ebf1f0051ff4d140cf7b292946bcc1a531b45e74/edx-remote-config/stage/lms.yml#L1832. Will update my PR to remove the extra underscore

learning_assistant/views.py
alangsto: I appreciate your response, and I think what you're saying makes a lot of sense! I like your above point about passing the enrollment mode to the audience, and letting optimizely decide whether or not to include the user in the experiment, as opposed to trying to filter out whether or not an optimizely decision should be made for a user on the frontend/backend. 

If including the enrollment in the user context for this PR still sounds like the right approach, I'll include that in this PR. Thank you again for your very thoughtful feedback 😊

learning_assistant/views.py
```diff
@@ -410,6 +410,6 @@ def get(self, request, course_run_id):
 
         data['audit_trial'] = trial_data
 
-        data['audit_trial_length_days'] = get_audit_trial_length_days()
+        data['audit_trial_length_days'] = get_audit_trial_length_days(user.id, enrollment_mode)
```
ilee2u, line 413: Is this redundant, if `get_audit_trial` is already going to contain the result of `get_audit_trial_length_days`?

learning_assistant/utils.py
```diff
@@ -123,19 +124,54 @@ def user_role_is_staff(role):
...
 
-    trial_length_days = getattr(settings, 'LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS', default_trial_length_days)
+    if (
+        variation['enabled']
+        and variation['variation_key'] == getattr(settings, 'OPTIMIZELY_LEARNING_ASSISTANT_TRIAL_VARIATION_KEY_28', '')
+    ):
+        trial_length_days = 28
+    else:
+        default_trial_length_days = 14
+        trial_length_days = getattr(settings, 'LEARNING_ASSISTANT_AUDIT_TRIAL_LENGTH_DAYS', default_trial_length_days)
```
ilee2u, line 174: A little confused by this logic, should we be gating whether or not a trial length of 14 days is returned on whether or not the experiment is `enabled` as well? My understanding was that only users in the experiment would get the audit trial, but lmk if that's incorrect or if I missed something (I know we went over gating on enrollment mode in standup as well).

ilee2u: This may also relate to Michael's comment above: https://github.com/edx/learning-assistant/pull/160#discussion_r1943545227. I'm a little lost on if the "control" variation is meant to also give you an audit trial of 14 days, and why we're doing that if that's the case.

learning_assistant/views.py
alangsto: Maybe? You'd have to derive the audit trial length based on the start and expiration date. This is also supposed to represent the setting itself. I could see this go either way

learning_assistant/utils.py
alangsto: If the experiment isn't enabled, the trial length should still be 14 days. The only case where the trial length should not be 14 days is if the experiment is enabled, and if the variation key is for the 28 day trial.

learning_assistant/views.py
ilee2u: Ahhh I see, you'd have to rearrange a few things to make that happen. I'll leave it up to you then whether to make a change on this or not.

ilee2u (approved):
LGTM 🚀

learning_assistant/utils.py
MichaelRoytman: I may have missed something, but why do we want to do that? We won't have any control to compare against, no? I thought we wanted to compare the following cases.

1. control: no audit experience
2. 14 day variation: 14 day audit trial experience
3. 28 day variation: 28 day audit trial experience

This matches [Varsha's discovery doc](https://2u-internal.atlassian.net/wiki/spaces/COSMO/pages/1630634111/Discovery+Design+the+Optimizely+Experiments+for+Xpert+LA+Audit+Experience), but let me know if I'm missing something.

If what you have currently is the design we want, shouldn't we combine `control` and the 14 day variation key? Otherwise, we're splitting a variation group into two, and I'm not sure why if they're seeing the exact same experience. But I think the intention was to follow what's in Varsha's discovery doc.

alangsto: If a user is in the control group, they won't see the Xpert audit experience in the first place, so what would we want the audit trial length to be in that case?

alangsto: The difference in the control and 14 day variation is that the control group will never have the Xpert UI accessible to them. Because of that, I don't see why I need to change what the backend returns, when they won't be interacting with the backend anyway.

MichaelRoytman: Ah, got it! I was coming at it from the perspective of what the correct value for this variable should be in the control group, which would be no audit trial at all. Would you mind dropping a short comment about that detail here?

alangsto: Added a comment to the code!

ilee2u: This is a note for myself mostly, but this is where the audit trial UI is gated in the frontend by the "14 day" and "28 day" keys: https://github.com/edx/frontend-lib-learning-assistant/blob/main/src/hooks/use-course-upgrade.js#L56-L58

PR #162, opened 2025-02-12
Add expiration_date to LearningAssistantAuditTrial model

MichaelRoytman:
**Description**

**Jira**: [COSMO-660](https://2u-internal.atlassian.net/browse/COSMO-660)

This commit adds an `expiration_date` field to the `LearningAssistantAuditTrial` model. This is an enhancement that makes it easier to know the exact expiration date of an audit trial. The audit trial length may change over time. For example, the setting may be changed, or we may experiment with different audit trial lengths. In those cases, it would challenging to know what the value of the length of the audit trial was when the learner created their audit trial. Adding this value to the model addresses this issue.


learning_assistant/utils.py
```diff
@@ -143,7 +143,7 @@ def get_optimizely_variation(user_id, enrollment_mode):
         variation_key = None
     else:
         optimizely_client = optimizely.Optimizely(sdk_key=settings.OPTIMIZELY_FULLSTACK_SDK_KEY)
-        user = optimizely_client.create_user_context(str(user_id), {'enrollment_mode': enrollment_mode})
+        user = optimizely_client.create_user_context(str(user_id), {'lms_enrollment_mode': enrollment_mode})
```
MichaelRoytman, line 146: This was changed to match Alie's changes in https://github.com/edx/frontend-lib-learning-assistant/pull/85.

tests/test_api.py
```diff
@@ -509,43 +509,50 @@ def test_expiration_date(
...
         self.assertEqual(expected_expiration_date, expiration_date)
 
 
+@ddt.ddt
 class GetAuditTrialTests(TestCase):
     """
     Test suite for get_audit_trial.
     """
 
-    @freeze_time('2024-01-01')
```
MichaelRoytman, line 521: This has no impact.

PR #171, opened 2025-03-04
Update version to 4.10.3 and update CHANGELOG.

michaelroytman:
### Description

This commit updates the library version to `4.10.3` and the CHANGELOG to include the most recent changes so that they can be published to PyPI.

**Jira**: None.


schenedx (approved):
👍

PR #172, opened 2025-03-04
Ilee2u/fix limit prompt template max length

ilee2u:
https://2u-internal.atlassian.net/browse/COSMO-672

### DIFF
--- a/learning_assistant/api.py
+++ b/learning_assistant/api.py
@@ -134,16 +134,13 @@ def render_prompt_template(request, user_id, course_run_id, unit_usage_key, cour
     template = Environment(loader=BaseLoader).from_string(template_string)
     data = template.render(unit_content=unit_content, skill_names=skill_names, title=title)
 
-    print("data length:", str(len(data)))
-    print("tokens used:", str(estimated_message_tokens(data)))
-
-    prompt_template_system_tokens = estimated_message_tokens(data)
-    max_tokens = getattr(settings, 'CHAT_COMPLETION_MAX_TOKENS', 16385)
-    remaining_tokens = max_tokens - prompt_template_system_tokens
-    print("remaining_tokens:", str(remaining_tokens))
-
-    token_limit = getattr(settings, 'LEARNING_ASSISTANT_AVAILABLE', False)
-    data = data[0:15000]
+    # The maximum of 15000 chars for the prompt template is based on average lengths
+    # of chat messages in production. This limit allows there to be a chat history of
+    # several 1000+ character long messages, with some room from buffer. This limit
+    # also prevents an error from occurring wherein unusually long prompt templates
+    # cause an error due to using too many tokens.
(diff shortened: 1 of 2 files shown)


learning_assistant/api.py
```diff
@@ -133,6 +133,14 @@ def render_prompt_template(request, user_id, course_run_id, unit_usage_key, cour
...
     template = Environment(loader=BaseLoader).from_string(template_string)
     data = template.render(unit_content=unit_content, skill_names=skill_names, title=title)
+
+    # The maximum of 15000 chars for the prompt template is based on average lengths
+    # of chat messages in production. This limit allows there to be a chat history of
+    # several 1000+ character long messages, with some room from buffer. This limit
+    # also prevents an error from occurring wherein unusually long prompt templates
+    # cause an error due to using too many tokens.
+    PROMPT_LENGTH_MAX = getattr(settings, 'CHAT_COMPLETION_PROMPT_TEMPLATE_MAX_CHAR_LENGTH', 15000)
+    data = data[0:PROMPT_LENGTH_MAX]
```
ilee2u, line 143: This is where I've set the max length of the prompt to 15000 characters

ilee2u:
Fun note: I played around with the prompt limit by inserting the entire script of the Bee Movie into a Unit. With a limit of 15,000 characters, only the first third of the movie's script is sent.

schenedx (approved):
👍

learning_assistant/utils.py
```diff
@@ -18,7 +18,7 @@
 log = logging.getLogger(__name__)
 
 
-def _estimated_message_tokens(message):
+def estimated_message_tokens(message):
```
michaelroytman, line 21: Why did you rename this? The underscore prefix typically indicates that the function is intended to be used as a private function. Python doesn't have a concept of public and private methods like Java does, so this is what we use to indicate to users that it shouldn't be used as a public function.

learning_assistant/api.py
```diff
@@ -133,6 +133,14 @@ def render_prompt_template(request, user_id, course_run_id, unit_usage_key, cour
 
     template = Environment(loader=BaseLoader).from_string(template_string)
     data = template.render(unit_content=unit_content, skill_names=skill_names, title=title)
+
+    # The maximum of 15000 chars for the prompt template is based on average lengths
```
michaelroytman, line 137: Can you be sure to use tokens here? If we use characters, we risk cutting words off. Tokens is the more appropriate concept to use here.

learning_assistant/utils.py
ilee2u: I had renamed this as I thought I had imported this into api.py in a previous commit, and I thought that meant I'd need to remove the underscore prefix.

learning_assistant/api.py
ilee2u: Trying to figure out how exactly to trim down the content of the message itself based on the number of tokens.

PR #173, opened 2025-03-05
Modify system prompt trimming logic and add user to raw_id_fields for LearningAssistantMessageAdmin form.

michaelroytman:
### Description

This pull request does the following...
* adds the user field of the `LearningAssistantMessage` model to the `raw_id_fields` of the `LearningAssistantAdmin` form to improve performance of the form
* modifies prompt trimming logic. Trimming the entire system prompt to the character limit will result in us trimming the actual static prompt text. What we want, instead, is to trim the unit content, which is the actual dynamic part of the system prompt.

**Jira**: [COSMO-672](https://2u-internal.atlassian.net/browse/COSMO-672) (private) for trimming logic, none for admin form.


schenedx (approved):
👍

PR #190, opened 2025-06-30
fix:  Ensure chat response is always returned as a list for XPert chat API compatibility

mraman-2U:
**Description:** 
This commit updates the chat response handling logic to ensure that responses from the XPert chat API are consistently returned as a list, regardless of whether the API returns a single object or a list. This change improves compatibility with the latest XPert chat API.

**Key changes:**

1. Modified learning_assistant/utils.py to wrap single object responses in a list.
2. Updated and expanded tests in tests/test_utils.py to verify correct handling for both single objects and list responses, as well as error cases.

**JIRA:**https://2u-internal.atlassian.net/browse/COSMO-700

### DIFF
--- a/learning_assistant/utils.py
+++ b/learning_assistant/utils.py
@@ -96,7 +96,11 @@ def get_chat_response(prompt_template, message_list):
                 data=json.dumps(body),
                 timeout=(connect_timeout, read_timeout)
             )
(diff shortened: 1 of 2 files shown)


ilee2u (approved):
LGTM 🚀

